PR #51, opened 2017-10-03
remove unused var in memberprofile.js

marshallhouston:
#### What does this PR do?
removes unused variable from MemberProfile.js


dviramontes (approved):
👍

Issue #55, opened 2017-10-13
List view of members


wdoug:
As a user, I would like an overview list of who is a member of the organization, so I can find members to view their profiles.

PR #60, opened 2017-10-13
54 list user emails

### LINKED ISSUE #54: Admin: View list of new user emails for adding to flowdock
As a Code for Denver admin onboarding new members, I would like an easy way to invite new members to Flowdock.

I believe the simplest way of achieving this initially is to put a list show a comma-separated list of emails of users (preferably filtered to be just those that signed up in the last week).

jksmall0631:
This PR closes #54 

#### What does this PR do?
-Modified members component to list all user emails


src/Members.js
```diff
@@ -1,34 +0,0 @@
-import React from "react";
-import gql from "graphql-tag";
-import { graphql } from "react-apollo";
-
-const Members = ({ members }) => {
-  return (
-    <div>
-      {members.map(member =>
-        <div key={member.email}>
-          {member.email}
```
wdoug, line 10: Interesting, we were already displaying users emails? Do we have other data for users such as their name?

src/UsersList.js
```diff
@@ -0,0 +1,35 @@
...
+  );
+};
+
+UsersList.defaultProps = {
+  users: []
+};
+
+const allUsersQuery = gql`
+  query users {
+    allUsers {
```
wdoug, line 23: Is there a way to filter by when users joined? Do we even store that data?

src/App.js
```diff
@@ -17,36 +17,35 @@ class App extends Component {
...
           exact path="/me"
-          render={() =>
-            <MemberProfile
-              user={this.props.data.user}
-            />}
+          render={() => <MemberProfile user={user} />}
         />
         <Route path="/me/edit" component={MemberProfileEdit} />
-      </div>
+        <Route path="/user-list" component={UsersList} />      </div>
```
wdoug: For this story, being a list of emails for the admin to add people, I think the route should be something like `/admin/onboarding`, and then card #55 could be the user list that would display more public info, like name, etc.

src/Members.js
jksmall0631: We were displaying Members before which was a seperate table in graph.cool. I just had to switch up the query. We can display name and email if you want, sure.

src/UsersList.js
jksmall0631: Yeah, we have a createdAt value. Looks like graph.cool will allow us to sort by this date or we can just create a query to grab only the people who signed up on that day. Let me know which you would like to do. @wdoug

src/App.js
jksmall0631: That sounds good to me, I'll make those changes now.

dviramontes (approved):
👍

dviramontes:
Made separate issues for the outstanding questions/features.

PR #64, opened 2017-10-31
removed node sass

Feldbot:
#### What does this PR do?
Removed Node-sass as it was causing lots of errors.


wdoug (commented):
@dviramontes is it cool if we do this? We were having tons of issues with node-sass (as I always do with projects that have it because of the compiled binaries) when trying to get this running on @Feldbot's computer.

If we want a compilation process, then I'd probably vote to use [postcss](https://github.com/postcss/postcss), which is purely javascript, and modular, so you can opt-in to different features, although I would also personally be interested in exploring a css-in-js solution (I've heard good things about [fela](https://github.com/rofrischmann/fela), but we'd probably want to discuss more).

dviramontes:
@wdoug yeah its cool by me. Sorry for the late response. Sorry to hear you all been having issues. I was planning on offering to help but yeah binaries are always tricky.

We don't have any `sass` files in the repo yet so its good timing. I would preferably not have write vanilla css / `postcss` and `fela` sound like good options to me, open to exploring other options as well. Lately i've been using http://tachyons.io/ because of the predefined classes.

dviramontes (approved):
👍

PR #68, opened 2017-11-28
53 profile view

### LINKED ISSUE #53: Add profile view
As a user, I would like to be able to see my profile and the profiles of other members in Code for Denver.

jbrennan414:
This PR closes #53 

#### What does this PR do?
Styles the profile page


src/App.css
```diff
@@ -1,9 +1,209 @@
...
+  padding: 5px;
+  border-right: 1px solid gray;
+}
+
+.contributed1 {
+  padding: 5px;
+  border-right: 1px solid gray;
+}
+
+.contributed2 {
```
nlkluth: take a look at :first-child, :nth-child, and :last-child. They might be able to help in this type of situation

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
-      <p>{user.email}</p>
-      <p>GitHub: {user.githubName}</p>
-      <p>Flowdock: {user.flowdockName}</p>
-      <Link to="/me/edit">edit</Link>
+  <MuiThemeProvider>
+    <div className="container">
+      <div className="card">
+        <div className="pic">
+          <div className="container-nested">
+            <img src="http://placehold.it/225x225"/>
```
wdoug: That's pretty nifty. Didn't know that was a thing

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
+          </div>
+        </div>
+        <div className="user-bio">
+          <div className="container-nested">
+            <h2>FirstName LastName</h2>
```
wdoug: Instead of using the `FirstName LastName` placeholder, we have the actual user's name already and can put it in here with `{user.name }`

src/MemberProfile.js
wdoug: This is the user's profile picture right? We can use the actual user picture data pulled from the account you sign up with:
```jsx
<img src={user.picture} alt="avatar" />
```

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
+            <p>"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
```
wdoug: Is this the user's description? We can put it in here with `{user.description}`

jbrennan414:
Hey guys, I'm seeing all these changes roll in, and I'm not certain of how to handle them. Is this something that I'll need to rectify before the branch is merged? How does this work from here?

Thanks! I appreciate all the feedback!

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
+        </div>
+        <div className="user-bio">
+          <div className="container-nested">
+            <h2>FirstName LastName</h2>
+            <p>"Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut 
+            labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris 
+            nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit 
+            esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in 
+            culpa qui officia deserunt mollit anim id est laborum."</p>
+            <a href="mailto:email@gmail.com">email@gmail.com</a>
```
wdoug: This can be the user's actual email with:
```jsx
<a href={user.email}>{user.email}</a>
```

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
+              <div className= "contact">
+                <span>GitHub: </span><a href="github.com">github.com</a>
```
wdoug: It looks like we also have the github name based on the code above this would be:
```jsx
<span>GitHub: </span><a href={`github.com/${user.githubName}`}>{user.githubName}</a>
```
Also do we need the span around the `GitHub` label? For that matter, I suppose an html `label` element would actually be more semantic if anything.

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
+              </div>
+              <div className= "contact">
+                <span>Flowdock: </span><a href= "flowdock.com">flowdock.com</a>
```
wdoug: It looks like `user.flowdockName` is a thing, so we can put that in here. I'm not sure if it'll make sense to link to something for that though. Maybe we should just have it be text. Maybe add an `@` in front of it?

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
+              </div>
+            </div>
+          </div>
+    <div className="user-experience">
+      <div className="contributions-nested">
+        <div className= "contributed">
+          <div>
+            <span className="hours-contributed">
+              1
```
wdoug: Hmm. I'm just realizing that while we might be able to get the hours contributed data from sparktime, we don't have that capability right now. The same goes for the other stats. What do we want to do about that? Do we want to hide this whole section until we can get those values, or do we want to have some sort of pending indicator that that feature hasn't been implemented yet?

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
+  </div>
+
+      <div className="card">
+        <div className="skillsleft">
+          <h2>Skills</h2>
+        </div>
+        <div className="skillsright"> 
+          <Chip className="skills-button" 
+          onRequestDelete= {()=> {return false}}>
+            Python
```
wdoug: Maybe we could put a pending indicator in here as well. Maybe we make it into a separate react component that is an ellipses (...) or something that when you mouse over it says: `Help us implement this feature`

src/MemberProfile.js
```diff
@@ -1,17 +1,115 @@
...
   return (
-    <div>
-      <img src={user.picture} alt="avatar" />
-      <h1>{user.name}</h1>
-      <p>{user.description}</p>
-      <p>{user.email}</p>
-      <p>GitHub: {user.githubName}</p>
-      <p>Flowdock: {user.flowdockName}</p>
-      <Link to="/me/edit">edit</Link>
+  <MuiThemeProvider>
```
wdoug, line 18: I believe we can actually pull this component up to wrap the entire app at the top of the `src/App.js` jsx. That way all our pages can use material design and take advantage of the theme we select

src/App.css
```diff
@@ -1,9 +1,209 @@
...
+  padding: 3px;
+  margin: 5px;
+}
+
+/* Be advised! The !important tag is to be used to override the default styling from 
+Material UI, without these !important tags, we won't be able to override Material UI's
+stupid inline style*/
+
+.skills-button {
+  font-family: "Roboto Condensed", sans-serif !important;
```
wdoug, line 190: Is the reason that we are using `!important` here because we are overriding inline styles on the elements themselves?

src/App.css
```diff
@@ -1,9 +1,209 @@
...
+
+.pic {
+  justify-content: center;
+  display: flex;
+  height: 400px;
+  flex-basis: 20%;
+  margin: 20px;
+}
+
+img {
```
wdoug, line 143: We definitely don't want this applied to all images. Also, as a heads up, in the world of web applications, I personally consider using html tag selectors in css directly generally bad practice as it can make the ability to track down css styles difficult once there starts to be lots of different styles

src/App.css
```diff
@@ -1,9 +1,209 @@
...
+  line-height: 40px;
+}
+
+.card {
+  display: flex;
+  background: white;
+  margin: 0 auto;
+  margin-bottom: 25px;
+  width: 90%;
+  box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
```
wdoug, line 129: `0.19`? Dang, that is so precise!

src/App.css
wdoug: Actually, I thought about this some more, and I suppose you'll want good sane defaults for basic things (e.g. headers, paragraphs, etc) defined directly with html tag selectors. I just don't think we want oval pictures to be the default style.

wdoug:
Hey @jpb55116, yes I think it would be great to get the suggestions that have been made updated before we merge the branch. These things are always a conversation though, so if there is a reason why we wouldn't want to address some of the comments, or why we'd want to have those changes in a separate pull request definitely let us know.

src/MemberProfile.js
jbrennan414: Not certain what you mean by this yet, for the time being, I'm going to handle it like the github username

src/App.css
jbrennan414: Yeah, Sean and I learned that Material-UI makes all of their styling in-line, so the only way for us to make changes was to use !important.

src/App.css
jbrennan414: Right?! Thanks Google!

src/App.css
dviramontes: Another way to do this without using `!important` would be to add another class to the element in question.
In CSS, specificity always wins:

`.skills-button.foo.bar { ... }`

dviramontes: Happy to help with this, just let me know! :)

src/MemberProfile.js
dviramontes: This is true but let's do it in a separate PR.

src/App.css
dviramontes: @wdoug 
@jpb55116 and I looked at this together and it looks like we need `!important` at least right now to overwrite the `<Chip/>` styles. There might be another way to do this but we can tackle this later. I'm ok with leaving it this.

wdoug:
This pull request is superceded by #78 which builds on top of this code

PR #90, opened 2018-01-05
Make travis build the client for both pull requests and master deploys

### LINKED ISSUE #86: ci: Make warnings have consistent failures on test and build
Right now warnings don't cause test failures but they cause build failures you can have a successful test on a pull request that then fails to deploy to production. Let's either make all warnings fail the build, or no warnings fail either testing or building.

wdoug:
This PR closes #86

#### What does this PR do?

- Runs the build step on both pull requests and on master. This is necessary so we don't have the build fail only on master.
- Adds caching of node_modules to speed up future builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,14 @@ language: node_js
 node_js:
   - 8.4.0
 
-script:
-  - "npm run lint"
-  - "npm test"
+cache:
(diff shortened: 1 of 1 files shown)


dviramontes (approved):
👍

PR #93, opened 2018-01-08
Changed auth0 setup to automate things a little more. 

### LINKED ISSUE #66: Improve onboarding with programmatic setup of Auth0 client
https://auth0.com/docs/api/management/v2

mugwump303:
Added 'bin/auth0setup' folder that contains all the files to be run by 'npm run setup'.
Updated 'package.json' file to run those files. 
Updated README file.

This PR closes #66 

#### What does this PR do?
"npm run setup" will now prompt the user for the **DOMAIN**, **CLIENT_ID**, AND **CLIENT_SECRET** of their API EXPLORER CLIENT.
If successful, it will create a '.env.local' file with **REACT_APP_AUTH0_CLIENT_ID**, **REACT_APP_AUTH0_DOMAIN**, and **REACT_APP_AUTH0_API_IDENTIFIER** filled out.

### DIFF
--- a/bin/auth0setup/create.js
+++ b/bin/auth0setup/create.js
@@ -1,0 +1,49 @@
+var envLocation = '.env.local'
+var dotenv = require('dotenv')
+dotenv.config({path: envLocation})
+
+var fs = require('fs')
+var request = require("request");
+var makeClient = require('./makeClient.js')
+
+
+// add an assert statement. double check that the env file exists. throw and error if it doesn't exist.
+
+
+// get a token
+var options = { method: 'POST',
+  url: 'https://' + process.env.DOMAIN + '/oauth/token',
+  headers: { 'content-type': 'application/json' },
+  body: 
+   { grant_type: 'client_credentials',
+     client_id: process.env.EXPLORER_CLIENT_ID,
+     client_secret: process.env.EXPLORER_CLIENT_SECRET,
+     audience: 'https://' + process.env.DOMAIN + '/api/v2/' 
+	 },
+	json: true 
+}
+
+
+request(options, function (error, response, body) {
+  
+	if(error) throw error; // TODO: add human readable error message for bad url/dns error.
+	
+	
+	// overwrite .env file with access token
+	var envFile = fs.readFileSync(envLocation, 'utf8')
+	var envJson = dotenv.parse(envFile)
+	envJson.ACCESS_TOKEN = body.access_token
+	var envString = ''
+	
+	for(var k in envJson) {
+		envString = envString + k + '=' + envJson[k] + '\n'
+	}
+	fs.writeFileSync(envLocation, envString)
+
+	// use the management api to create a new client
+  makeClient(body.access_token)
+
+});
+
+
+
--- a/bin/auth0setup/create.js
+++ b/bin/auth0setup/create.js
@@ -1,48 +1,51 @@
 var envLocation = '.env.local'
 var dotenv = require('dotenv')
-dotenv.config({path: envLocation})
 
 var fs = require('fs')
 var request = require('request');
-var makeClient = require('./makeClient.js')
 
 
 // add an assert statement. double check that the env file exists. throw and error if it doesn't exist.
 
 
-// get a token
-var options = { method: 'POST',
-  url: 'https://' + process.env.DOMAIN + '/oauth/token',
-  headers: { 'content-type': 'application/json' },
-  body: { 
-    grant_type: 'client_credentials',
-    client_id: process.env.EXPLORER_CLIENT_ID,
-    client_secret: process.env.EXPLORER_CLIENT_SECRET,
-    audience: 'https://' + process.env.DOMAIN + '/api/v2/'
-  },
-  json: true 
+var create = function() {
+	
+	var makeClient = require('./makeClient.js')
+	dotenv.config({path: envLocation})
+
+	// get a token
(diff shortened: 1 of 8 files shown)


bin/auth0setup/promptUser.js
```diff
@@ -0,0 +1,35 @@
...
+
+fs.writeFileSync('.env.local', '')
+
+prompt.start();
+
+prompt.get(['DOMAIN', 'EXPLORER_CLIENT_ID', 'EXPLORER_CLIENT_SECRET'], function (err, result) {
+  if (err) { return onErr(err); }
+
+  // overwrite .env file with access token
+  var envFile = fs.readFileSync('.env.local', 'utf8')
```
wdoug: 😕 Here we're reading from a file that we are writing to on [line 6](https://github.com/codefordenver/members/pull/93/files#diff-f096c3631bef8aaea21378db9b0425c5R6) to be an empty file?

wdoug: Is this supposed to be reading from the `.sample-env` file?

package.json
```diff
@@ -9,13 +9,16 @@
   "dependencies": {
     "apollo-client": "^1.9.0",
     "auth0-js": "^8.11.2",
+    "dotenv": "^4.0.0",
```
wdoug: These three added dependencies are only required for development so they should live in `devDependencies` in this file.

bin/auth0setup/promptUser.js
```diff
@@ -0,0 +1,35 @@
+var fs = require('fs')
+var dotenv = require('dotenv')
+dotenv.config()
+var prompt = require('prompt');
+
+fs.writeFileSync('.env.local', '')
```
wdoug: If users already have a `.env.local` file, can we warn them that they are about to override it and prompt them to continue before we override it?

bin/auth0setup/promptUser.js
```diff
@@ -0,0 +1,35 @@
...
+  envJson.DOMAIN = result.DOMAIN
+  envJson.EXPLORER_CLIENT_ID = result.EXPLORER_CLIENT_ID
+  envJson.EXPLORER_CLIENT_SECRET = result.EXPLORER_CLIENT_SECRET
+
+  var envString = ''
+	
+  for(var k in envJson) {
+    envString = envString + k + '=' + envJson[k] + '\n'
+  }
+  fs.writeFileSync('.env.local', envString)
```
wdoug: Instead of writing these values to this file temporarily and having another npm script run by `npm run setup` that pulls in these values, can we instead just call the create script directly from this script with the values from the prompt instead? I think that could greatly simplify what is happening here.

wdoug (dismissed):
I had an issue when I tried running this, possibly based on a recent change to the environment variables 😬 

Maybe we can chat about it at some point

wdoug:
I really like what you've got started though @mugwump303! This is going to be way better!

bin/auth0setup/create.js
```diff
@@ -0,0 +1,49 @@
...
+  // overwrite .env file with access token
+  var envFile = fs.readFileSync(envLocation, 'utf8')
+  var envJson = dotenv.parse(envFile)
+  envJson.ACCESS_TOKEN = body.access_token
+  var envString = ''
+	
+  for(var k in envJson) {
+    envString = envString + k + '=' + envJson[k] + '\n'
+  }
+  fs.writeFileSync(envLocation, envString)
```
wdoug: @mugwump303 would you feel comfortable refactoring this code to collect all the variables from the created Auth0 APIs first, and then only write to the envFile once at the very end? I think this could be done really nicely if we use [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) (maybe using [request-promise-native](https://github.com/request/request-promise-native), or maybe even just using something like node's [promisify utility](http://2ality.com/2017/05/util-promisify.html)) and [async await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function). I believe it could be done with something along these lines:

Changing requests like the one made in `makeClient` to have an exported function that returns a promise for the value(s) we care about. 
```js
// makeClient.js
const requestPromise = require('request-promise-native');
// ...
module.exports = async function makeClient(token) {
  // ...
  const response = await requestPromise(options);
  return response.client_id;
}
```

Then the final version of this file could look something like:
```js
async function setUpAuth0andSaveEnvironmentVariables(/* prompted variables */) {
  try {
    const token = await getAuth0Token(domain, explorerClientId, explorerClientSecret);
    const clientId = await makeClient(token);
    const apiIdentifier = await makeAPI(token);
    saveVariablesToEnvFile({
      REACT_APP_AUTH0_DOMAIN: domain,
      REACT_APP_AUTH0_CLIENT_ID: clientId,
      REACT_APP_AUTH0_API_IDENTIFIER: apiIdentifier,
    });
  } catch (err) {
    console.error('There was an error with the automated setup of auth0 for this project:');
    console.error(err);
    process.exit(1); // not sure if this is necessary or not
  }
}
// prompt for initial Auth0 keys to work with
setUpAuth0andSaveEnvironmentVariables(/* prompted variables */);
```

mugwump303: That might take me a little more time, so I'll try to fix a few of the other things while looking into Promises and Aysnc/Await.

fredmon3 (approved):
Looks great! :shipit:

PR #105, opened 2018-01-23
#96 add date-fns email list

### LINKED ISSUE #96: feat: display members who joined in the last week
As a Code for Denver admin, I would like the `EmailList` component to only display emails of members who joined in the last week so I can easily invite them to flowdock.

sungbae:
This PR closes #96

Added date-fns library to easily calculate newest members in the last week.

### DIFF
--- a/src/EmailList.js
+++ b/src/EmailList.js
@@ -30,7 +30,7 @@ const allUsersQuery = gql`
 const EmailListWithData = graphql(allUsersQuery, {
   options: {
     variables: {
-      date: new Date(format(subDays(new Date(), 7), 'YYYY-MM-DDTHH:mm:ss.SSSZ'))
+      date: format(subDays(new Date(), 7), 'YYYY-MM-DDTHH:mm:ss.SSSZ')
     }
   },
   props: ({ data: { allUsers } }) => ({ users: allUsers })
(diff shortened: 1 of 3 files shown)


src/EmailList.js
```diff
@@ -19,17 +20,19 @@ EmailList.defaultProps = {
...
-  }
-`;
+    } 
+  }`;
 
 const EmailListWithData = graphql(allUsersQuery, {
-  options: () => ({}),
+  options: {
+    variables: {
+      date: new Date(format(subDays(new Date(), 7), 'YYYY-MM-DDTHH:mm:ss.SSSZ'))
```
wdoug: Is it necessary to wrap the `format` with a `new Date`? That seems odd...

sungbae: Right, this is because graphql is expecting a date type of DateTime (at least, that's what I think). format() will spit out the ISO string so I wrapped it in new Date() to translate that as a DateTime type. I can double check if this is necessary later today!

sungbae: @wdoug good catch, this works with just the ISO string of the date. I'll get rid of this new Date().

wdoug (approved):
:shipit:

PR #119, opened 2018-01-30
fix src pathto pick up the right

### LINKED ISSUE #115: Fix the logo on pages other than the landing page
Should look like:

Instead, on any page besides the landing page if you load that page directly it looks like:

Probably just need to set `img` `src` to have a leading slash e.g. `` instead of ``

KeynesYouDigIt:
This PR closes #115

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,6 +14,8 @@ import MemberProfilePage from './MemberProfilePage';
 class App extends Component {
   render() {
     const { User } = this.props.data || {};
+    const CfD_logo_location =
+      process.env.PUBLIC_URL + '/images/cfd-circle-white.png';
     return (
       <MuiThemeProvider>
         <div className="App">
@@ -21,7 +23,7 @@ class App extends Component {
             <a href="/">
               <img
                 className="cfd-logo"
-                src="../images/cfd-circle-white.png"
+                src={CfD_logo_location}
                 alt="code for denver logo"
               />
             </a>
--- a/src/App.js
+++ b/src/App.js
@@ -21,7 +21,7 @@ class App extends Component {
             <a href="/">
               <img
                 className="cfd-logo"
-                src="images/cfd-circle-white.png"
+                src="../images/cfd-circle-white.png"
                 alt="code for denver logo"
               />
             </a>


KeynesYouDigIt:
Not a front end guy so if this is a not-ok hack

KeynesYouDigIt:
for #115

src/App.js
```diff
@@ -21,7 +21,7 @@ class App extends Component {
             <a href="/">
               <img
                 className="cfd-logo"
-                src="images/cfd-circle-white.png"
+                src="../images/cfd-circle-white.png"
```
sungbae: perfectly fine as long as we keep directory structure consistent.

sungbae: I just took a look at out current folder structure and it seems like you would need to change that line to: `"../public/images/cfd-circle-white.png"`

So curious how the current change is working... Let me know if this makes sense.

wdoug: @KeynesYouDigIt this doesn't work for routes that are more than 2 slashes deep. See [/a/a/a](https://deploy-preview-119--cfd-members.netlify.com/a/a/a) as an example. The way that `create-react-app` is set up builds all the static files in `public` into the root of what gets served. This means that instead of using `../` you can just use `/images/cfd-circle-white.png` and it should work for all the cases.

wdoug (changes requested):
If you change `../images/cfd-circle-white.png` to `../images/cfd-circle-white.png` that should cover all the cases. See my other comment for an explanation

wdoug:
@KeynesYouDigIt I also just added you to the @codefordenver/members-projects-team so you can have write access to the repo and open pull requests directly from branches rather than forks. We've found that that generally makes things easier to collaborate on.

src/App.js
wdoug: Actually, the other option for this, that `create-react-app` recommends it to utilize webpack and `import` the image directly into the javascript file. [See this for info](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-images-fonts-and-files)

wdoug: [More info on when to use the `public folder`](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#when-to-use-the-public-folder)

KeynesYouDigIt: Well derp. youre right. updates pending.........

src/App.js
```diff
@@ -14,14 +14,16 @@ import MemberProfilePage from './MemberProfilePage';
 class App extends Component {
   render() {
     const { User } = this.props.data || {};
+    const CfD_logo_location =
+      process.env.PUBLIC_URL + '/images/cfd-circle-white.png';
     return (
```
KeynesYouDigIt, line 19: https://github.com/codefordenver/members/blob/master/docs/CreateReactApp.md line 684 seems to imply this is the best way to reference the public folder. LMK if thats not the case.

KeynesYouDigIt:
@wdoug Should be ready for another looksee

src/App.js
KeynesYouDigIt: quick note - when I use the import statement I cant get it to compile.

```
./src/App.js
Module not found: You attempted to import ../public/images/cfd-circle-white.png which falls outside of the project src/ directory. Relative imports outside of src/ are not supported. You can either move it inside src/, or add a symlink to it from project's node_modules/.
```

if I attempt to make the import absolute (this must not do that)

```
Module not found: Can't resolve 'public/images/cfd-circle-white.png' in '/home/vince/Dropbox/TZU/CfD/members/src'
```
Then with the leading slash
```
./src/App.js
Module not found: You attempted to import /public/images/cfd-circle-white.png which falls outside of the project src/ directory. Relative imports outside of src/ are not supported. You can either move it inside src/, or add a symlink to it from project's node_modules/.
```

wdoug: That's interesting about the `PUBLIC_URL` param.

@KeynesYouDigIt in pull request #124 I added an `images` directory in `src` (we might want to rename it to `assets` or something), you could follow that pattern. That way we can import with the relative path (see [here](https://github.com/codefordenver/members/pull/124/files#diff-1ab3942ea02626f4ecee3ae12ed527d0R2)).

wdoug: I think we want to use the relative import, not just because that is recommended by create react app, but because the way create-react-app is set up, using the relative import will automatically include it in the service worker so it works offline.

wdoug:
This is superseded by: #158

PR #122, opened 2018-02-02
84 made card lenght fixed

### LINKED ISSUE #84: Fix responsiveness of profile page
Right now when the profile page is viewed in a narrow browser the styling is broken. Let's make it properly responsive.

ghost:
This PR closes #84

#### What does this PR do?
Elements now resize on the profile page when the window resizes.


src/App.css
```diff
@@ -130,22 +131,27 @@ a {
   height: 100%;
   padding: 20px;
   align-items: center;
+  min-width: 1024px;
```
wdoug: To me, responsive design means that content is reorganized on smaller screens so users don't have to horizontally scroll (unless maybe on screens even narrower than the narrow view from a cell phone). Looking at the chrome devtools [device mode](https://developers.google.com/web/tools/chrome-devtools/device-mode/) shows that the screen width of the Galaxy S5 is 360 pixels, which is much smaller than 1024, and we might want to support even thinner screens.

wdoug: Also, we now have the newer version of the material-ui library, which has built in [flexbox layout support](https://material-ui-next.com/layout/grid/) that we may want to use.

src/App.css
```diff
@@ -8,6 +8,16 @@ html {
   box-sizing: inherit;
 }
 
+.App-header {
```
wdoug, line 11: We don't need this anymore because of changes from #132

wdoug (approved):
@justinktang, one comment, then :shipit:

PR #125, opened 2018-02-06
106 ensure node version

### LINKED ISSUE #106: Ensure proper npm version before `npm install` and `npm start`
There is a specific version of npm and node for the application. The version of node to make sure the features are the same in all environments. The version of npm is important because it defines the format of the `package-lock.json` file. 

As a developer, I want to have these versions enforced so I don't see unexpected results and huge changes to the package-lock file for no reason.

evanplaice:
This PR closes #106

#### What does this PR do?

- adds semver as a devDependency
- adds /src/util/checkVersion.js
- adds a 'check-version' npm script
- runs a version check against the engine version specified in `package.json`
  - before `npm install`
  - before `npm start`


evanplaice:
CI doesn't want to install semver. Going to implement the version check without it.

evanplaice:
Netlify CI fails because the node version specified in the config needs to be >= 8.9.0.

8.9.3 is the latest LTS release.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 8.4.0
+  - 8.9.3
```
wdoug, line 3: Let's update the `.nvmrc` file as well so people using nvm will have the correct version.

package.json
```diff
@@ -28,13 +27,15 @@
...
+    "start": "npm run check-version && react-scripts start",
     "build": "react-scripts build && cpy build/index.html build --rename 404.html",
     "lint": "eslint .",
+    "preinstall": "npm run check-version",
     "postinstall": "npm install --prefix server",
     "setup": "node ./bin/auth0setup/promptUser.js",
     "gc": "cd ./server && dotenv -e ../.env.local graphcool",
-    "deployProdBackend": "cd ./server && graphcool deploy"
+    "deployProdBackend": "cd ./server && graphcool deploy",
+    "check-version": "node ./src/utils/checkVersion.js"
```
wdoug: I would probably put this in the `bin` directory as a script that is associated with setup, whereas the way this repo is set up the `src` directory is where the client code lives.

evanplaice: Good point, not sure how I missed /bin before. Isn't the usual convention to put CLI scripts in /.bin?

wdoug:
I just came across this [check-node-version](https://www.npmjs.com/package/check-node-version) package. Would that simplify things for us?

evanplaice:
I originally used semver to do the check. For some reason CI wouldn't install the package. That package has 7 top-level deps incl semver. IMO, it's better to avoid piling packages onto the dep tree for such a simple use case.

The current implementation will ensure greater-than or greater-than-or-equal. Anything else and it'll throw an error.

evanplaice:
ESLint is complaining about the console statements. Is there any special way I should log messages to the console, or should I just add an exception to this script?

wdoug:
Holy moly, 7 dependencies? In that case, I'm cool to keep this approach.

I would just disable the `no-console` rule for the bin directory. 🤔 ...or maybe we should just disable that from the whole repo... either seems fine to me.

evanplaice:
The no-console rule makes sense for client code. You wouldn't want a forgotten debug statement spitting output to the console in the background. CLI scripts are the obvious exception because that's the only way to provide feedback to the user.

bin/checkVersion.js
```diff
@@ -0,0 +1,48 @@
+/*eslint no-console: "off"*/
```
wdoug, line 1: I think we could set this for the whole `bin` directory (https://github.com/codefordenver/members/blob/master/bin/.eslintrc), but this is fine for now.

wdoug (approved):
:shipit:

wdoug:
@evanplaice you may squash and merge when ready

evanplaice:
OK, squashed and ready to go!

PR #128, opened 2018-02-09
Add initial graphql schema type for projects

wdoug:
#### What does this PR do?

Adds an initial graphql type to graphcool for projects. This can easily be extended in the future, but is going to be more complicated if we need to migrate things so let's discuss different use cases now before we merge this.

#### Please Check
- typos
- naming is intuitive and makes sense
- flexible enough to support future use cases

### DIFF
--- a/server/types.graphql
+++ b/server/types.graphql
@@ -32,3 +32,19 @@ type User @model {
 #   # Every relation also required a back-relation (to determine 1:1, 1:n or n:m)
 #   author: User! @relation(name: "UserPosts")
 # }
+
+type Project @model {
+  # Required system field:
+  id: ID! @isUnique # read-only (managed by Graphcool)
+
+  # Optional system fields (remove if not needed):
+  createdAt: DateTime! # read-only (managed by Graphcool)
+  updatedAt: DateTime! # read-only (managed by Graphcool)
+
+  name: String!
+  description: String
+  repo: String #url
+  flow: String #url
+  waffleBoard: String #url
+  driveDirectory: String #url
+}


server/types.graphql
```diff
@@ -32,3 +32,19 @@ type User @model {
...
+  # Required system field:
+  id: ID! @isUnique # read-only (managed by Graphcool)
+
+  # Optional system fields (remove if not needed):
+  createdAt: DateTime! # read-only (managed by Graphcool)
+  updatedAt: DateTime! # read-only (managed by Graphcool)
+
+  name: String!
+  description: String
+  repo: String #url
```
wdoug, line 46: My initial thought with these is that each of these options could potentially be plural (e.g. multiple associated repos), however, I personally think that case is very uncommon, and if we run into it we could either add additional fields to support that, or simply have projects describe those cases in the `description` field.

dviramontes: Agreed. Owlet is the only one that comes to mind.

dviramontes (approved):
👍

evanplaice (commented):
If you're looking for a more extensible approach.

- keep the name and description fields static
- make the repo a list of repositories
- repositories contains a [name, url]
- repositories can be extended later if necessary to include additional fields (ex contributors)
- replace flow/waffleBoard/driveDirectory with a 'resource' field
- resources contain 2 fields, [name, url]

 As long as the lists don't become really long lookup time shouldn't be an issue.

It may be beneficial to include additional extensions in the future. Such as tagging, donors lists, etc.

wdoug:
I've been thinking about this a bit, and I'm not sure if we know enough at this point whether it is more beneficial to have things like `repo` be changed into just a more specifically named `repoUrl` string field, or a relationship to a separate `repo` or `resource` type. I think I'm going to start with a pull request that just adds the `name` and `description` fields. We can do quite a bit on the client with those. 
-- cc: @dviramontes @evanplaice

wdoug:
Superceded by #139

PR #129, opened 2018-02-09
This connects to #72

mugwump303:
Connects to #72


src/Auth.js
```diff
@@ -1,13 +1,15 @@
 const ACCESS_TOKEN_KEY = 'access_token';
 const EXPIRES_AT_KEY = 'expires_at';
 const USER_ID = 'user_id';
+const BEARER_TOKEN = 'cfd-members-auth0IdToken';
```
mugwump303, line 4: There was already a line in index.js that takes the bearer token in localStorage and uses it in the Authorization header. However, I couldn't find a place where the bearer token was actually set, so I put it here.

server/src/auth0/updateUser.graphql
```diff
@@ -0,0 +1,3 @@
+query canUserUpdate($user_id: ID!, $input_id: ID!) {
+  SomeUserExists(filter: { AND: [{ id: $user_id }, { id: $input_id }] })
```
wdoug, line 2: What is the `$input_id`?

server/graphcool.yml
```diff
@@ -34,8 +34,11 @@ functions:
 # Read more here:
 # https://docs-next.graph.cool/reference/auth/authorization/overview-iegoo0heez
 permissions:
-- operation: "*"
-
+# - operation: "*"
+  - operation: User.read
+  - operation: User.update
+    authenticated: true
+    query: ./src/auth0/updateUser.graphql
```
wdoug, line 41: At this point, what does this query get us over just `authenticated: true`?

wdoug:
I'm guessing we'll probably want a follow up to this work to add admin roles that can delete and update other users.

server/src/auth0/updateUser.graphql
mugwump303: `$input_id ` is the `id` argument of the mutation.

server/graphcool.yml
mugwump303: This query makes sure that the User sending the mutation (`$user_id`) is the User being modified (`$input_id`).

wdoug (changes requested):
I'm seeing some weird issues on this branch (I don't know if it just needs to be rebased off master or if there is something else).

First, when logging in, the nav-bar still shows the login button instead of the user's profile link.

Second, when I try to edit myself on `/me/edit`, I get the following error:
```error
MemberProfileEdit.js:52 Uncaught TypeError: Cannot read property 'id' of null
    at MemberProfileEdit.updateDB (MemberProfileEdit.js:52)
    at onClick (MemberProfileEdit.js:88)
    at HTMLUnknownElement.callCallback (react-dom.development.js:542)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:581)
    at Object.invokeGuardedCallback (react-dom.development.js:438)
    at Object.invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:452)
    at executeDispatch (react-dom.development.js:836)
    at executeDispatchesInOrder (react-dom.development.js:858)
    at executeDispatchesAndRelease (react-dom.development.js:956)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:967)
    at Array.forEach (<anonymous>)
    at forEachAccumulated (react-dom.development.js:935)
    at processEventQueue (react-dom.development.js:1112)
    at runEventQueueInBatch (react-dom.development.js:3607)
    at handleTopLevel (react-dom.development.js:3616)
    at handleTopLevelImpl (react-dom.development.js:3347)
    at batchedUpdates (react-dom.development.js:11082)
    at batchedUpdates (react-dom.development.js:2330)
    at dispatchEvent (react-dom.development.js:3421)
```

When I tried updating another user directly through Graphql I did see the expected error message "Insufficient permissions for this mutation", so that is good. Not sure what is happening with the other issues.

mugwump303:
I wasn't able to get the error you got.
I tried to do `git rebase master` then `git push origin 72-user-permissions`, but I got an error when trying to push to github.
Do you think it'd be easier to just delete this branch, recreate it from the current master, then apply the few changes there?

wdoug (approved):

PR #146, opened 2018-02-24
Add ability to create and edit projects

### LINKED ISSUE #141: Add ability to create projects

### LINKED ISSUE #143: Add ability to edit projects

wdoug:
This PR closes #143, closes #141 

#### What does this PR do?

- Adds the ability to edit projects
- Adds the ability to create projects
- Adds higher order components for wrapping view, edit, and create pages


wdoug:
I'm not super happy with the implementation of this right now. I'd like to refactor things so there is less boilerplate. I'm not sure exactly what path I want to go down for that though. Maybe adding [redux-form](https://redux-form.com) -- on one hand it seems a little silly to add redux just for forms, but on the other hand, it also opens up the door to using some redux ecosystem stuff, which could possibly make it easy to do things like persisting partially edited forms to local storage.

wdoug:
Alright, I've refactored things, and while I think we can continue to make things better, they are way better than they started out, and I feel generally comfortable merging things as they stand now

src/sections/ProjectSection.js
```diff
@@ -1,18 +1,26 @@
...
-  }
-  const { name, description } = project;
+import EditableText from '../forms/EditableText';
+import EditableMarkdown from '../forms/EditableMarkdown';
 
+const ProjectSection = ({ formData, editing, onFormDataChange }) => {
+  const { name, description } = formData;
+  const commonProps = {
+    onChange: onFormDataChange,
+    editing
```
sungbae, line 9: Is this supposed to be a property? What does this do exactly?

src/sections/ProjectSection.js
```diff
@@ -1,18 +1,26 @@
...
+  };
   return (
-    <div>
-      <h1>{name}</h1>
-      <p>{description}</p>
-    </div>
+    <React.Fragment>
+      <h1>
+        <EditableText value={name} label="Title" name="name" {...commonProps} />
```
sungbae, line 14: I guess you're injecting the elements into here (e.g. `{...commonProps}`)? But I still don't understand what `editing` is exactly...

src/utils/withCreatePage.js
```diff
@@ -0,0 +1,83 @@
...
+          this.setState({ submitting: false });
+          alert(err);
+        }
+      };
+
+      _updateFormData = event => {
+        const { name, value } = event.target;
+        this.setState(state => ({
+          formData: {
+            ...state.formData,
```
sungbae, line 73: Is this how you can copy/paste the properties from `state.formData` here?

src/utils/withEditPage.js
```diff
@@ -0,0 +1,93 @@
...
+      };
+
+      componentWillReceiveProps(nextProps) {
+        if (nextProps.formData !== this.props.formData) {
+          this.setState({ formData: nextProps.formData });
+        }
+      }
+
+      render() {
+        if (!this.props.formData) {
```
sungbae, line 37: Couldn't this always be false? If `this.props.formData` is initialized as `{}`.

src/utils/withViewPage.js
```diff
@@ -0,0 +1,44 @@
...
+ * @returns (Component) => WrappedComponent
+ */
+export default function withViewPage(options) {
+  return WrappedComponent => {
+    class ViewPageWrapper extends React.PureComponent {
+      static displayName = `WithViewPage(${getDisplayName(WrappedComponent)})`;
+
+      render() {
+        const { formData } = this.props;
+        if (!formData) {
```
sungbae, line 21: Same concern as the comment in `withEditPage.js`. If the `formData` is initialized already as `{}`, couldn't this always be false?

sungbae (changes requested):
I think this is pretty solid. I just left a few questions!

src/sections/ProjectSection.js
wdoug: It is shorthand for `editing: editing` in the object

src/utils/withCreatePage.js
wdoug: It is a way of copying the keys and values from `formData` into this new object

src/utils/withEditPage.js
wdoug: It is passed in as a prop from above, which, as it is coded currently, initializes it as `undefined`

sungbae (approved):

rufusasterisk:
Tested on the deploy/netlify link

##### Add new event
Adding a new event from /projects/create successfully created an event - link was visible on /projects
The event persisted after navigating away
The event body supports markdown

##### Edit existing event
Event saved changes made through edit process
Changes persisted after navigating away

Issue #127, opened 2018-02-09
Don't require page reload on signup/login


wdoug:
Right now when a user signs up or logs in the page does a full refresh to show them as logged in. This seems to sometimes create issues, and is very slow.

mugwump303:
I think the cause is the "window.location.reload()" call in the LoginAuth0 component. 
We probably need to pass the new user data into one of the parent components, then the Login button should be re-rendered with the correct user data without a full page reload.

I think we can use "refetch", "update", or "refetchQueries"

mugwump303:
I'm not sure if the way I ended up doing it is good. 

Here are some links to Apollo docs that looked promising. I couldn't get them to work though.

https://www.apollographql.com/docs/react/basics/setup.html#compose
https://www.apollographql.com/docs/react/basics/mutations.html#graphql-mutation-options-update
https://www.apollographql.com/docs/react/basics/mutations.html#graphql-mutation-options-refetchQueries

PR #175, opened 2018-03-16
connects to 127 - fix reloading issue

mugwump303:
This PR closes #[127]
connects to #127

#### What does this PR do?
Uses new user data to update components without reloading page.


server/.eslintrc
```diff
@@ -8,4 +8,4 @@
   "rules": {
     "no-console": "off"
   }
-}
+}
```
wdoug, line 11: Why're we removing the newline at the end of this file?

mugwump303: Not sure. I didn't even think about it when I pasted it back in. I can put it back in.

wdoug (changes requested):
I'm not crazy about this solution. The underlying Auth system that we have right now isn't set up to work properly with React which is forcing us to use hacks like `forceUpdate`. I'm going to try to refactor the Auth stuff to work better with React, which should allow us to manage this better and set us up better for the future. Hopefully that will go smoothly. If not, I will approve this and we can merge it.

wdoug:
This PR is superseded by #178

PR #178, opened 2018-03-22
Refactor auth management and remove extra reloads on login/logout

### LINKED ISSUE #127: Don't require page reload on signup/login

### LINKED ISSUE #161: Fix netlify server paths
Currently netlify test server will reroute to the CFD website rather than itself.

eg. member.codefordenver.org/volunteers not deploy-cfd-members.netlify.com/volunteers

We want the latter. 

To reproduce go to any pull request, expand "Show all checks" in the auto testing section, and click on "details" of the deploy/netlify section.The initial page is netlify but after login it redirects to CFD.

wdoug:
This PR closes #127 - Overrides #175 

#### What does this PR do?

Refactors the Auth management to 
- get it to work better with React
- support session syncing across tabs
- removes unnecessary reloading on login/logout (#127)
- fix redirect url for netlify (closes #161)


wdoug:
You can test the results of this change [here](https://deploy-preview-178--cfd-members.netlify.com/)

src/sections/LoginAuth0.js
```diff
@@ -23,38 +23,42 @@ class LoginAuth0 extends Component {
...
   componentDidMount() {
-    const props = this.props;
     this.webAuth.parseHash(
       { hash: window.location.hash },
       (err, authResult) => {
+        window.location.hash = '';
         if (err) {
-          // TODO: Handle errors
+          // TODO: Handle errors better
+          alert(`${err.error}: ${err.errorDescription}`);
```
object88, line 38: Don't forget to remove this.

src/sections/LoginAuth0.js
```diff
@@ -23,38 +23,42 @@ class LoginAuth0 extends Component {
...
           return console.error(err);
         }
         if (!authResult || !authResult.accessToken) {
           return;
         }
-        window.location.hash = '';
+        this.setState({ isLoggingIn: true });
```
object88, line 44: Consider using the [callback](https://reactjs.org/docs/react-component.html#setstate) on `setState`, instead of passing in a hash.  I think this way of assigning state is old practice.

(Same on line 61 below.)

src/utils/withAuthSession.js
```diff
@@ -0,0 +1,114 @@
...
+  return {
+    authSession: getAuthSession(),
+    isAuthenticated: isAuthenticated()
+  };
+}
+
+const componentsWithAuth = [];
+
+function handleChange() {
+  componentsWithAuth.forEach(comp => comp.setState(getUpdatedState()));
```
object88, line 64: See notes on `setState` above.

src/sections/LoginAuth0.js
object88: (Or is not there another error reporting system?)

src/sections/LoginAuth0.js
dviramontes: @object88 The callback seems like an overkill if we are updating a single value. I think perfectly fine to use a hash in this case.

dviramontes (approved):
LGTM

src/sections/LoginAuth0.js
wdoug: I believe that only matters for state updates that are dependent on the current state since `setState` is asynchronous. 
> "If the next state depends on the previous state, we recommend using the updater function form, instead"

-- https://reactjs.org/docs/react-component.html#setstate

However, it might be best to just be consistent everywhere. That way we encourage people to use functions which work in all cases instead of objects which sometimes might not work properly. I'll make the change

src/sections/LoginAuth0.js
wdoug: Not yet. The `alert` is there because it is better than having a silent failure.

src/sections/LoginAuth0.js
wdoug: Actually, I might just push those changes out into a separate piece of work since there are several other cases of `setState` using objects, and if I merge this as is, I don't have to re-test things (which is all manual right now since there are essentially no automated tests for this project currently).

PR #180, opened 2018-03-27
Add some suggested steps for new members to help streamline onboarding

wdoug:
This is really just a temporary step before we make a better onboarding experience for new members


dviramontes (approved):
👍

PR #187, opened 2018-03-30
added ProjectChampions relation that connects User and Project types

### LINKED ISSUE #162: Add champions field to projects
As a member, I want to be able to see who I can contact as a champion of a project. This could also allow us to have more fine-grained permissions for managing the project on the site in a nice way. 

This should support multiple Users as champions

mugwump303:
This PR closes #162 

Adds ProjectChampion relation that connects two types (User and Project).


server/types.graphql
```diff
@@ -17,6 +17,7 @@ type User @model {
   flowdockName: String
   description: String
   role: UserRole @defaultValue(value: REGULAR)
+  projects: Project @relation(name: "ProjectChampions")
```
wdoug: We want a many-to-many relationship here since members can have multiple projects they are championing/have championed. I think we might also want to have this called `projectsChampioned` or something more specific since we could have other relationships to projects from users as well.

wdoug:
I just went to test this, and it looks like we need to explicity add the permission to do this so we can call `addToProjectChampions`

wdoug (changes requested):
I get this error when trying to edit a project as an admin:

screen shot 2018-04-19 at 8 44 27 pm

wdoug:
I wonder if we want to temporarily keep allowing all users to update projects. That makes it so we can merge this pull request without having to make any changes to the front-end

wdoug (approved):

PR #189, opened 2018-04-03
Node Version Checking Fix

### LINKED ISSUE #185: Node Version Checking Fix
Currently, v8.11.0 doesn't pass our node version testing, even though it is a newer version than 8.9.3

mugwump303:
This PR closes #185 

#### What does this PR do?
Fixes node.js version checking by changing array of strings to array of numbers.

### DIFF
--- a/bin/checkVersion.js
+++ b/bin/checkVersion.js
@@ -9,6 +9,14 @@ const version = config.engines.node.match(/[0-9.]+/).toString();
 var min = minVersion.split('.');
 var ver = version.split('.');
 
+// turn elements in array from string to number
+min = min.map(function(value, index) {
+  return Number(value);
+});
+ver = ver.map(function(value, index) {
+  return Number(value);
+});
+
 // version comparator
 var test = ((a, b, comp) => {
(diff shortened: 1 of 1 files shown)


bin/checkVersion.js
```diff
@@ -9,6 +9,14 @@ const version = config.engines.node.match(/[0-9.]+/).toString();
 var min = minVersion.split('.');
 var ver = version.split('.');
 
+// turn elements in array from string to number
+min = min.map(function(value, index) {
```
wdoug, line 13: These maps could be done on the same line as `min` is originally defined i.e.
```js
var min = minVersion.split('.').map(v => Number(v));
```

wdoug (approved):
:shipit:

bin/checkVersion.js
mugwump303: Should I keep it as is, or make the change?

wdoug: Nah. I'd say leave it as is and go ahead and merge it @mugwump303

PR #209, opened 2018-04-27
Add routes for user who is not logged in and corresponding test

japadgett2:
This PR closes no active issue, it only allows this developer to contribute for a first time with something he felt comfortable doing in React

#### What does this PR do?
It directs a user who isn't logged in to our current log in page when they try to access an internal page (currently: /projects, /me, /volunteers)

#### Feedback
Feedback is very welcome.

### DIFF
--- a/src/__tests__/routing.test.js
+++ b/src/__tests__/routing.test.js
@@ -4,9 +4,39 @@ import App from '../App';
 
 describe('when not logged in', () => {
   describe('when trying to navigate to /volunteers', () => {
-    it('shows the 404 page', () => {
+    it('shows the landing page w/prompt to log in', () => {
       const wrapper = mountWithContext(<App />, {
         routes: ['/volunteers']
+      });
+
+      expect(wrapper.contains('Please sign up or log in')).toBe(true);
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


src/__tests__/routing.test.js
```diff
@@ -4,11 +4,41 @@ import App from '../App';
...
       const wrapper = mountWithContext(<App />, {
         routes: ['/volunteers']
       });
 
+      expect(wrapper.contains('Please sign up or log in')).toBe(true);
+    });
+  });
+
+  describe('when trying to navigate to /projects', () => {
+    it('shows the landing page w/prompt to log in', () => {
```
wdoug, line 17: Adding tests? Woohoo!

wdoug (approved):
I'm cool with this. One other thought that I had was to make the logged-out 404 page have login prompt (similar to GitHub's 404 page if you're not logged in). One other thing that would be nice would be to have the user go back to/stay on the page they were on when they log (back) in, which, now that I think about it, I think would be easier to implement with just updating the not-logged-in 404 page to prompt a login. @japadgett2

japadgett2:
Wow, I missed this one.  Thanks for assigning.  I will look into this.

PR #210, opened 2018-04-27
193 dropdown heading

### LINKED ISSUE #193: Make user section in top bar match design
- [ ] Dropdown profile with logout option
- [ ] Option to go to profile view
- [ ] Make sure this looks decent in mobile

Design:

jberns:
This PR closes #193 

#### What does this PR do?
This PR adds a drop down menu for the Profile and Logout Buttons.
Menu.js ->  This file handles the new heading.  The avatar, username, and logout functionality are linked in from the old Login.js file.
Header.js -> Updated the Grid spacing to look better on mobile.  Changed the link styles to buttons to be consistent with the design.


wdoug (changes requested):
Yeah boi! Look at you gettin' proficient at all this React work!

My one requested change is to not have the header take up so much screen real estate on small screens (aka mobile). Right now, for my user on an iPhone 5 the header takes up half the screen.

See:
screen shot 2018-05-17 at 6 22 15 pm

In contrast, here is what it looks like on master right now:
screen shot 2018-05-17 at 6 23 42 pm

src/sections/Header.js
```diff
@@ -14,27 +15,48 @@ import './Header.css';
...
+          {isAuthenticated && (
+            <Grid container justify="flex-start" alignItems="center">
+              <Grid item sx={3}>
+                <Link className="Header-link" to="/volunteers">
+                  <Button color="secondary">ALL USERS</Button>
+                </Link>
+              </Grid>
+              <Grid item sx={3}>
+                <Link className="Header-link" to="/projects">
+                  <Button color="secondary">ALL PROJECTS</Button>
```
wdoug: Let's instead use the `component` prop on these to make it a link for better accessibility and less markup. See https://material-ui-next.com/demos/buttons/#third-party-routing-library

src/sections/Menu.js
```diff
@@ -0,0 +1,113 @@
...
+import Paper from 'material-ui/Paper';
+import { MenuItem, MenuList } from 'material-ui/Menu';
+import { withStyles } from 'material-ui/styles';
+import ClickAwayListener from 'material-ui/utils/ClickAwayListener';
+import { Link } from 'react-router-dom';
+
+import './Menu.css';
+
+const styles = theme => ({
+  root: {
```
wdoug: Nifty

src/sections/Menu.js
```diff
@@ -0,0 +1,113 @@
...
+    display: 'flex'
+  },
+  paper: {
```
wdoug: I don't think this is used

src/sections/Menu.js
```diff
@@ -0,0 +1,113 @@
...
+              </Button>
+            </div>
+          </Target>
+          <Popper
+            placement="bottom"
+            eventsEnabled={open}
+            className={classNames({ [classes.popperClose]: !open })}
+          >
+            <ClickAwayListener onClickAway={this.handleClose}>
+              <Grow
```
wdoug: Interesting!

src/sections/Menu.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { Manager, Target, Popper } from 'react-popper';
```
wdoug: We already have `react-popper` in this project? Curious

src/sections/Menu.js
```diff
@@ -0,0 +1,113 @@
...
+  popperClose: {
+    pointerEvents: 'none'
+  }
+});
+
+class MenuListComposition extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      avatar: props.avatar,
```
wdoug: No need to put these on state, you can get them directly off props.

src/sections/Menu.js
```diff
@@ -0,0 +1,113 @@
...
+  },
+  paper: {
+    marginRight: theme.spacing.unit * 2
+  },
+  popperClose: {
+    pointerEvents: 'none'
+  }
+});
+
+class MenuListComposition extends React.Component {
```
wdoug: I think this is more complicated than we need, and I'm a fan of just using the simpler material-ui components when they exist instead of integrating with React-popper. I personally find the [change-transition example](https://material-ui-next.com/demos/menus/#change-transition) a lot simpler to reason about and probably a lot simpler to maintain.

src/sections/Menu.js
```diff
@@ -0,0 +1,113 @@
...
+            <ClickAwayListener onClickAway={this.handleClose}>
+              <Grow
+                in={open}
+                id="menu-list-grow"
+                style={{ transformOrigin: '0 0 0' }}
+              >
+                <Paper>
+                  <MenuList role="menu">
+                    <Link to="/me" style={{ textDecoration: 'none' }}>
+                      <MenuItem onClick={this.handleClose}>PROFILE</MenuItem>
```
wdoug: It looks like [MenuItem's also take a `component` prop](https://material-ui-next.com/api/menu-item/) so we can pass in `Link` to that instead of wrapping it.

src/sections/Menu.js
jberns: This is no longer used.

src/sections/Menu.js
jberns: Hm.  I'm not sure if this is something I had installed locally when I was setting up the material-ui library on my comptuer as it does not show up in the package.json file...  However I changed the menu type so this is no longer applicable.

src/sections/Menu.js
jberns: I updated the menu to [change-transition example](https://material-ui-next.com/demos/menus/#change-transition).  However for some reason I could not import the 'material-ui/Fade' transition component locally but the menu still functions without it.

src/sections/Menu.js
jberns: Updated the MenuItem's component prop to use the Link component.  Neat trick!

src/sections/Header.js
jberns: I changed the buttons back to Links from the master branch,  The buttons have different spacing than the link components which is making the header much bigger.

src/sections/Menu.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import Avatar from 'material-ui/Avatar';
+import Icon from 'material-ui/Icon';
+import Button from 'material-ui/Button';
+import { MenuItem } from 'material-ui/Menu';
+import { Link } from 'react-router-dom';
+import Menu from 'material-ui/Menu';
+/*import Fade from 'material-ui/Fade';*/
```
wdoug: Any reason to keep this line?

wdoug (approved):
LGTM

PR #224, opened 2018-06-05
214 cfapi projects

### LINKED ISSUE #214: [spike] Use cfapi to pull data for repos tagged "active-project" as the source of truth for projects
Get:
- [ ] Repo name, description, url
- [ ] Issues labeled "help wanted" or "good first issue"

fredmon3:
This PR closes #214

#### What does this PR do?
This PR uses the Code For America API to retrieve project information.


package.json
```diff
@@ -6,6 +6,7 @@
     "node": ">=8.9.0"
   },
   "dependencies": {
+    "@material-ui/core": "^1.2.0",
```
wdoug: I see this is on top of your other PR

wdoug:
The idea that I had was to still be able to edit the metadata associated with the project in this project, so basically simultaneously query the graphql project, except query it by `cfaProjectId` instead of just `id`. If that is the case I guess we'd just remove the `/projects/create` page, and if a user tries to edit a project that we don't yet have data for we'd want to create it on the graphql side.

src/pages/ProjectsListPage.js
```diff
@@ -14,11 +14,15 @@ const ProjectsList = ({ projects, loading }) => {
...
-        <li key={project.id}>
-          <Link to={`/projects/${project.id}`}>{project.name}</Link>
-        </li>
-      ))}
+      {projects
+        .filter(project => project.status === 'Ideation')
+        .map(project => (
+          <li key={project.id}>
+            <Link to={`/projects/${project.id}`}>
+              {project.description || project.name}
```
wdoug: It seems like it should be the project name, followed by the project description to me

wdoug:
@fredmon3 see my comments in [this flowdock thread](https://www.flowdock.com/app/cfa-brigades/members-project/threads/D1MlcAgrRWgTMYTtqI_yeFUQPK4) I don't think the CfA API will work for us. Maybe just using github's rest api directly is a better option... or we could talk about other options again

fredmon3:
@wdoug We can close this PR then

PR #236, opened 2018-07-17
Alert user on duplicate email address (#233)

### LINKED ISSUE #233: Gracefully handle user email duplication
Right now, the email field in the graphcool is schema is marked as unique. When a pre-existing user signs in via different connection but with a duplicate email, the insertion into the database will fail and is not handled.

dchao19:
This PR closes #233 

#### What does this PR do?
This PR prevents users who have already signed in with the members website to sign in with a different social connector with a same email. This will cause errors in the database and previously, the authentication request will fail silently without notifying the user the request was unsuccessful. 

This PR adds the `id_token` scope to the Auth0 request so the client can receive basic user profile information, and removes the `userinfo` Auth0 Authentication API call because the decoded id_token contains the profile information we were using.. 

This PR removes `isomorphic-fetch` as a dependency in the server.


server/src/auth0/auth0Authentication.js
```diff
@@ -1,140 +1,136 @@
-const fetch = require('isomorphic-fetch');
```
wdoug, line 1: I wonder why git thinks this whole file has changed. I hope it isn't line-ending changes. Good thing for being able to filter out whitespace changes with `w=1` in the url.

src/sections/LoginAuth0.js
```diff
@@ -49,16 +59,32 @@ class LoginAuth0 extends Component {
...
+              picture
             }
           })
           .then(({ data }) => {
             const userInfo = data.authenticateUser;
             this.props.setAuthSession(authResult, userInfo.id, userInfo.token);
             this.setState({ isLoggingIn: false });
+          })
+          .catch(error => {
+            const alertMessage = error.message.includes(
```
wdoug: We should have a polyfill for this for [users using internet explorer](https://caniuse.com/#search=includes). It doesn't look like [our version of CreateReactApp polyfills it](https://github.com/codefordenver/members/blob/master/docs/CreateReactApp.md#supported-language-features-and-polyfills), and, while I thought we had the babel-polyfill included, I don't see it.

src/sections/LoginAuth0.js
```diff
@@ -49,16 +59,32 @@ class LoginAuth0 extends Component {
...
-              accessToken: authResult.accessToken
+              accessToken: authResult.accessToken,
+              name,
+              email,
+              picture
             }
           })
           .then(({ data }) => {
             const userInfo = data.authenticateUser;
+            this.setState({ isLoggingIn: false });
```
wdoug, line 75: I changed the order of this and the following line to fix this warning:
screen shot 2018-07-19 at 6 29 37 pm

wdoug:
I went ahead and added the `babel-polyfill` so [users on internet explorer can have the `String.prototype.includes` method](https://caniuse.com/#search=includes).

wdoug (approved):
LGTM. @dchao19 check my changes that I made and if they look good, I'd say :shipit:

dchao19:
They were definitely line ending problems. Sorry about that. The diff is fixed now.

Rest of the changes look good. `babel-polyfill` is probably a good addition either way.

PR #240, opened 2018-07-20
Query for projects by repoName

wdoug:
#### What does this PR do?

- Swaps the url to use the Project `repoName` instead of `id`, which makes for prettier urls
- Shows a warning on the page when the issues fail to load:
screen shot 2018-07-19 at 9 22 12 pm

### DIFF
--- a/server/types.graphql
+++ b/server/types.graphql
@@ -48,5 +48,5 @@ type Project @model {
   name: String! @isUnique
   description: String
   cfapiProjectId: String @isUnique @depricated # Soon to be removed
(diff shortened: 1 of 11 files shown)


wdoug:
I suppose there are some tradeoffs to the url change. This forces us to have github repos for projects, which I believe has been the case for everything we've done in the past, but I'm not sure if it is necessarily going to be the case in the future. Thoughts?

wdoug:
I'm going to hold off on this for now. I don't know if forcing projects to have an associated repo is the direction we want to go. We might want to have new project pitches get created here first before any repo is created for them

PR #250, opened 2018-10-02
Add skills to projects

### LINKED ISSUE #249: Add editable skills for projects

wdoug:
This PR closes #249 

#### What does this PR do?

Adds editable skills to projects


src/sections/ProjectSection.js
```diff
@@ -34,4 +45,19 @@ const ProjectSection = ({ project, editing, onFormDataChange }) => {
   );
 };
 
+ProjectSection.fragments = {
```
dchao19, line 48: These are really nice - maybe at some point we should consider moving these into their own query file as the data is used throughout the application.

src/sections/ProjectSection.js
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
+import gql from 'graphql-tag';
 import EditableText from '../forms/EditableText';
 import EditableMarkdown from '../forms/EditableMarkdown';
+import EditableSkills from '../forms/EditableSkills';
 import ProjectIssues from './ProjectIssues';
 
 const ProjectSection = ({ project, editing, onFormDataChange }) => {
```
dchao19, line 8: This whole edit project section needs some styling, but the functionality is still there and it doesn't undermine this PR I don't think

dchao19 (approved):
Everything looks good! Definitely some things to tidy up in the future though, but its nice to finally start somewhere. Thanks!

src/sections/ProjectSection.js
wdoug: I definitely think we can use fragments a lot better. I'm not sure how I feel about putting them in their own query files though as that seems to have less of the original benefit of graphql of co-locating queries next to components.

PR #251, opened 2018-10-04
Add a skills detail page with associated users and projects

wdoug:
NOTE: Depends on PR #250 

#### What does this PR do?

- Add a skills detail page with associated users and projects
- Links to this page from the skill chips on project and user pages
- Renames the EditableChips component into an EditableList component with and makes it more reusable


dchao19 (approved):
Does this PR also need the schema updates from #250?

wdoug:
This PR was built on top of #250, so that should be in already in here

PR #257, opened 2018-10-16
Fix website in microsoft edge via fetch-headers polyfill

### LINKED ISSUE #255: Website fails to load on Microsoft Edge
Using versions: 

Microsoft Edge 42.17134.1.0
Microsoft EdgeHTML 17.17134

When browsing to members.codefordenver.org, a plain white screen is displayed without any content. 
An error is printed to the console: `SCRIPT65535: Invalid argument`

dchao19:
This PR closes #255 

#### What does this PR do?
This PR adds the fetch-headers polyfill to fix Apollo-link-rest in Microsoft Edge. This isn't an ideal solution because it polyfills regardless of browser support, but Microsoft Edge has the wrong fetch-headers and it is incompatible with Apollo-link-rest. For more details: see https://github.com/apollographql/apollo-link-rest/issues/93.

Importantly, however, it fixes the website in Microsoft Edge.

### DIFF
--- a/src/config/polyfills.js
+++ b/src/config/polyfills.js
@@ -1,3 +1,12 @@
-global.Headers = require('fetch-headers');
+import Headers from 'fetch-headers';
+
 // This is necessary for IE / Microsoft Edge because of a problem with apollo-link-rest
 // https://github.com/apollographql/apollo-link-rest/issues/93
+window.Headers = Headers;
+
+// The polyfill doesn't provide a forEach function which apollo rest uses
+window.Headers.prototype.forEach = function(fn) {
+  for (let [key, value] of this.entries()) {
+    fn(value, key, this);
+  }
+};
--- a/src/config/polyfills.js
+++ b/src/config/polyfills.js
@@ -1,12 +1,1 @@
-import Headers from 'fetch-headers';
-
-// This is necessary for IE / Microsoft Edge because of a problem with apollo-link-rest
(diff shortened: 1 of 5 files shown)


src/config/polyfills.js
```diff
@@ -0,0 +1,3 @@
+global.Headers = require('fetch-headers');
```
wdoug: It's a little odd to me to be assigning this to `global` and just using the fact that our webpack setup has aliased `global` to `window` to be compatible with node

wdoug:
I'm going to go ahead and rebase this to fix the package-lock.json conflict and then merge it

wdoug:
I updated this branch, and I also submitted a pull request against the fetch-headers library to fix a few issues with it: https://github.com/jimmywarting/fetch-headers/pull/4

I think we should use [this polyfill](https://github.com/github/fetch) instead and just use their implementation of the Headers class.

src/config/polyfills.js
dchao19: Switching the polyfill to whatwg-fetch resolves this comment.

dchao19:
You're right. I've updated this PR to use the whatwg-fetch polyfill and the fix appears to have stuck on Edge with versions:

Microsoft Edge 42.17134.1.0
Microsoft EdgeHTML 17.17134

wdoug (approved):
LGTM

PR #271, opened 2018-11-13
Update checkVersion to use compare-versions and remove preinstall script

### LINKED ISSUE #270: Fix checkVersion
**Description of Issue:**
The checkVersion.js script is incorrectly failing when using node version 10.13.0.  The script is reporting the following error:  Node version 10.13.0 not >= 10.8.0

image

**Expected Behavior:**
I would expect the script to report a success for any node versions greater than what is specified in the package.json file.

**Proposed Fix:**
Update checkVersion.js to use [compare-versions](https://github.com/omichelsen/compare-versions)
Remove preinstall script from package.json and only apply check-version to npm start

cvanem:
This PR closes #270

#### What does this PR do?
Modifies checkVersion.js to use [compare-versions](https://github.com/omichelsen/compare-versions) and correctly compares/validates the process node version against the version listed in package.json.

It also modifies the package.json file and removes the preinstall script as compare-versions needs to be installed prior to executing.


cvanem:
Failed CI tests and unintentional modifications to package-lock.json.  Will re-submit new PR.

PR #272, opened 2018-11-14
Update checkVersion to use compare-versions and remove preinstall script

### LINKED ISSUE #270: Fix checkVersion

cvanem:
This PR closes #270

#### What does this PR do?
Adds compare-versions to package.json and package-lock.json.

Modifies checkVersion.js to use [compare-versions](https://github.com/omichelsen/compare-versions) and correctly compares/validates the process node version against the version listed in package.json.

It also modifies the package.json file and removes the preinstall script as compare-versions needs to be installed prior to executing.  

**If validation pre npm install is a requirement, then we can could add another package.json file to the bin directory with just the check-version dependency.  Let me know if you want to do this as I didn't want to pollute the bin directory without someone else looking at this implementation.**


wdoug:
I had originally suggested doing something like this, but the whole point of this script is to run in the `preinstall` to make sure you have the correct environment before installing dependencies. 

I see that the logic in that file is not correct, but I'd propose we just fix it and keep it from needing a dependency

cvanem:
@wdoug
Ok, we could do that.  I will strip out the necessary logic from compare-versions and put it in checkVersion.js if that works?  It is fairly small.

bin/checkVersion.js
```diff
@@ -1,6 +1,7 @@
 /*To avoid dependencies, the compareVersions code below has been copied from: https://github.com/omichelsen/compare-versions*/
 /*eslint no-console: "off"*/
 /*eslint no-useless-escape: "off"*/
+/*eslint indent: "off"*/
```
wdoug, line 4: Is this rule even enabled?

bin/checkVersion.js
```diff
@@ -1,41 +1,100 @@
...
-  }
-  return true;
-}
+const version = process.version.match(/[0-9.]+/).toString();
+const minVersion = config.engines.node.match(/[0-9.]+/).toString();
+const semver = /^v?(?:\d+)(\.(?:[x*]|\d+)(\.(?:[x*]|\d+)(\.(?:[x*]|\d+))?(?:-[\da-z\-]+(?:\.[\da-z\-]+)*)?(?:\+[\da-z\-]+(?:\.[\da-z\-]+)*)?)?)?$/i;
+
+const comparators = [
+  { val: '>', validate: r => r == 1 },
+  { val: '>=', validate: r => r == 0 || r == 1 }
```
wdoug, line 13: I don't think we care to limit this to just these to comparators, but this is fine for now

cvanem: Ok, I can go ahead and add the other ones in. It won't be much more effort

wdoug: Actually, I don't see this code handling the logic for `^` or `~` semver options so this is good

wdoug (approved):
:shipit:

bin/checkVersion.js
cvanem: Ok

bin/checkVersion.js
cvanem: I may have had that enabled locally...Let me remove it.  Do you want me to delete this branch and submit a new PR so everything is a little cleaner?

cvanem: Apparently it is enabled as the CI is complaining of the format.  I tried to fix the actual code, but it gets re-formatted on commit.

Issue #111, opened 2018-01-26
Add docs for suggesting apollo dev tools for development


wdoug:
Apollo has some sweet dev tools. Let's note that in the readme!

wdoug:
Depends on #137

PR #275, opened 2018-11-15
Typescript spike

### LINKED ISSUE #262: [Spike] Evaluate adding typescript
create-react-app 2 [now supports typescript](https://github.com/facebook/create-react-app/pull/4837)

Depends on #266

wdoug:
This PR closes #262 

#### What does this PR do?

This pull request investigates adding [Typescript](https://www.typescriptlang.org/) to this project so we can get the benefits that typescript provides. Before I go much further on this investigation, I figured I should start the conversation around if this is something we want to do.

Adds [graphql-code-generator](https://graphql-code-generator.com/) to automatically generate typescript types and apollo HOCs and Query and Mutation components directly from graphql queries.

### Benefits of this setup:
- Type checking catching type errors => has already helped me catch some errors
- Ability to have more clarity around what data you are dealing with (e.g. prop types)
  - This can also help with working with library APIs (I noticed we were incorrectly using react-apollo for example)
- Nice editor integrations (e.g. autocomplete on types)
- Ability to generate types from the graphql schema and queries, thus allowing end-to-end type safety

### Downsides
- The generics for typing the react-apollo `graphql` higher order component can be really confusing to deal with -> Much better with the new graphql-code-generator setup
- ~~Currently you'd have to manually remember to update those types~~ - Fixed with the new graphql-code-generator setup
- Create React App won't render the client if there are type errors even if the code would actually work which can be frustrating (we should look into if there is a setting to turn this off)
- Some of the material-ui typing (especially around `withStyles` and themes) are annoying and I haven't gotten around to figuring out how to handle them

### Other notes:
- apollo codegen doesn't work if multiple versions of the `graphql` dependency are installed
  - ~~In order to get it working I had to use Yarn~~ - switched back to npm for now
  - ~~Also I ended up using an older version of apollo because I was having issues with the newer one (it might have just been a configuration issue though)~~ - swapped with graphql-code-generator


cvanem:
Typescript is a must have for me.  It can be a pain at first, but I find it very useful in other projects I have worked on.   A couple notes:

It will integrate nicely if we migrate from Graphcool to Prisma (Typescript).

Typescript interfaces are mostly optional, so existing code should still function without them.  You can get around compilation errors that would otherwise work by specifying an any type.  

I'm fairly experienced with the Material-UI typings and how they work if you need any assistance.

I think your biggest downside is going to be the learning curve and syntax adjustments.

Also, I prefer yarn to npm, so that is a plus for me.

cvanem:
You will probably also want to change react-scripts to react-scripts-ts

wdoug:
@cvanem in regard to switching from `react-scripts` to `react-scripts-ts`, now that create-react-app uses [babel 7 that supports typescript](https://babeljs.io/blog/2018/08/27/7.0.0#typescript-support-babel-preset-typescript) it [supports typescript directly](https://github.com/facebook/create-react-app/pull/4837) and is now the recommended way to use typescript with create-react-app (see the note about switching from create-react-app-typescript [here](https://facebook.github.io/create-react-app/docs/adding-typescript#docsNav))

Awesome, I may reach out to you for help with the Material-UI typing.

cvanem:
@wdoug Good to know.  I guess I can update my other projects now :)

src/authenticated/member/MemberProfilePage.tsx
```diff
@@ -0,0 +1,53 @@
...
+    };
+  };
+};
+type ComponentProps = {
+  user?: getUser_user;
+  editing: boolean;
+  onFormDataChange: (value: any) => void;
+};
+
+const MemberProfilePage = graphql<
```
dchao19: Ooph - those typings are...verbose. I understand why we use recompose and the graphql function, but this seems like a broader architectural concern about the overuse of the "withAuthSession" HoC / helper in so many of the components. the <Query> component + render function seems like the recommended pattern if we don't use recompose. Perhaps @wdoug your concerns about the complex generics of the graphql function can be alleviated if we switch to Query component as much as possible. See typing for the Query component [here](https://www.apollographql.com/docs/react/recipes/static-typing.html#typing-components).

dchao19:
I love the developer experience with Typescript - there's a lot less flipping tabs to make sure types and objects are accurate, and some of your downsides should be expected when using TS. I personally think it's a better idea for CRA to block rendering unless the type checking succeeds, but this will definitely be an adjustment for developers unfamiliar with typescript. 

I agree with @cvanem that we can make this a gradual conversion - i.e require new components be written in Typescript and slowly work on upgrading and refactoring old ones as they are used.

src/authenticated/member/MemberProfilePage.tsx
wdoug: Yeah, good point! It would definitely be worth (at the very least investigating) moving to the Query component. I was also wondering if we could wait for a React hooks implementation for Apollo (see: https://www.npmjs.com/package/react-apollo-hooks), but it seems like there are some [fundamental limitations in the current `useContext` api](https://twitter.com/acemarke/status/1055694323847651335) that may delay that for a little while.

 I also totally agree about issues with the `withAuthSession` HOC. It would be really nice to refactor that.

wdoug:
Another question on this PR is: 
If we do want to move to Typescript, what else would we want to do on this PR before merging it?

I think we'd definitely want to do the following:
- [x] Clean up the way that type generation works and make it not hard-coded to my local graphcool deployment
- [x] Update the docs
- [ ] ~~Clean up situation with yarn (if that is what we are going to) (e.g. delete package-lock.json), and inforce its use in the `preinstall` script~~ - Switched back to just using npm for now

Nice to haves:
- [x] Investigate what typing is like for the `Query` and `Mutation` apollo components and see how they compare to the Higher Order Component typings
- [x] Investigate doing type generation with the [graphql-code-generator] library instead and what we'd have to refactor for that use case (see [example](https://github.com/dotansimha/graphql-code-generator/tree/master/examples/typescript-react-apollo))

What else? Are there other "high risk" files that we'd want to implement typings for?

cvanem:
@wdoug Just curious if you noticed any differences in webpack compile times compared to the JS configuration?  There seem to be a few CRA issues reported that indicate it is slow, but I'm guessing that may only be noticeable on larger projects.  

I attempted to convert one of my other projects and did notice that there are a few options in the tsconfig.json that aren't supported yet.    Most notably, the ability to specify import paths from your root directory and re-importing modules (CRA requires isolatedModules=true).  This is probably a non-issue for this project.

Also, I am not sure how much you have looked at the tsconfig.json file yet, but you can adjust the settings in that file to loosen the typescript checks if you need to turn off some of the errors.

wdoug:
I didn't notice compile times being slower, but that is probably because of how small this project is. 

Cool. I haven't looked into the tsconfig much yet.

wdoug:
@cvanem, @dchao19 I swapped out the typescript code generation from graphql with [graphql-code-generator](https://graphql-code-generator.com/) and it feels waaayy better now. Mind looking at the current setup?

wdoug:
- [x] One other thing that would be nice to fix if possible is getting eslint working with `.graphql` files

wdoug:
I have updated the readme and now it closes #111 as well

server/graphcool.yml
```diff
@@ -77,6 +77,6 @@ permissions:
 # Your root tokens used for functions to get full access to the API
 # Read more here:
 # https://docs-next.graph.cool/reference/auth/authentication/authentication-tokens-eip7ahqu5o
-# rootTokens:
-# - mytoken
+rootTokens:
+- ${env:GRAPHQL_ROOT_TOKEN_SECRET}
```
wdoug: I don't think this is needed

server/src/.eslintrc
```diff
@@ -0,0 +1,3 @@
+{
+  "plugins": ["graphql"]
```
wdoug, line 2: This allows us to avoid throwing syntax errors on graphql files in this directory. In the future we could maybe add some graphql lint rules here, but would need to use [this eslint plugin](https://github.com/cjoudrey/graphql-schema-linter) instead

dchao19 (approved):
LGTM!

PR #277, opened 2018-11-25
Setup project champions UI

### LINKED ISSUE #195: Add UI for setting and viewing project champions
Projects can have champions. Let's update the UI to show them. Designs can be seen [here](https://www.figma.com/file/dHAyvZXkSvRW4nC5Q7JhoD3g/code-for-denver-members-project-1)

Depends on #162

cvanem:
This PR closes #195 

Adds a user select interface for project champions under the project profile.
Adds a project select interface for selecting projects under the user profile.

Please confirm the desired behavior.  

1. Under the user profile, the user is allowed to select ~~and create new~~ projects.
2. Under the project profile, users (aka project champions) can only be selected, not created.

Also, please verify that the CSS is acceptable.  I wasn't sure if I should match the wire frame designs exactly or if just matching the existing style is acceptable?

### DIFF
--- a/src/authenticated/member/MemberProfile.js
+++ b/src/authenticated/member/MemberProfile.js
@@ -4,7 +4,7 @@ import gql from 'graphql-tag';
 import HelpUsImplementThis from '../../shared-components/HelpUsImplementThis';
 import LoadingIndicator from '../../shared-components/LoadingIndicator';
 import EditableSkills from '../../forms/EditableSkills';
-import EditableList from '../../forms/EditableList';
+import EditableProjects from '../../forms/EditableProjects';
 import EditableText from '../../forms/EditableText';
 import EditableMarkdown from '../../forms/EditableMarkdown';
 import './Member.css';
@@ -20,7 +20,8 @@ const MemberProfile = ({ user, onFormDataChange, editing }) => {
     githubName,
     flowdockName,
     email,
-    skills
+    skills,
+    projectsChampioned
   } = user;
   const commonProps = {
     onChange: onFormDataChange,
@@ -113,14 +114,14 @@ const MemberProfile = ({ user, onFormDataChange, editing }) => {
         <div className="Member-skills-left">
           <h2>Projects</h2>
         </div>
-        <HelpUsImplementThis>
-          <div className="Member-skills-right">
-            <EditableList
-              value={[{ name: 'Members Project' }, { name: 'Owlet' }]}
-              label="Add Project"
-            />
-          </div>
-        </HelpUsImplementThis>
+        <div className="Member-skills-right">
+          <EditableProjects
+            value={projectsChampioned}
+            name="projectsChampioned"
+            label="Add Project"
+            {...commonProps}
+          />
+        </div>
       </Card>
     </div>
   );
@@ -129,6 +130,14 @@ MemberProfile.fragments = {
   skills: gql`
     fragment Skills on User {
       skills {
+        id
+        name
+      }
+    }
+  `,
+  projects: gql`
+    fragment Projects on User {
+      projectsChampioned {
         id
         name
       }
--- a/src/authenticated/member/MemberProfile.js
+++ b/src/authenticated/member/MemberProfile.js
@@ -112,7 +112,7 @@ const MemberProfile = ({ user, onFormDataChange, editing }) => {
 
       <Card className="Member-card">
         <div className="Member-skills-left">
-          <h2>Projects</h2>
+          <h2>Projects Championed</h2>
         </div>
         <div className="Member-skills-right">
           <EditableProjects
--- a/src/authenticated/member/MemberProfileEditPage.js
+++ b/src/authenticated/member/MemberProfileEditPage.js
@@ -12,6 +12,7 @@ const updateUserQuery = gql`
     $flowdockName: String
     $description: String
     $skillsIds: [ID!]
+    $projectsChampionedIds: [ID!]
   ) {
     updateUser(
       id: $id
@@ -20,6 +21,7 @@ const updateUserQuery = gql`
       flowdockName: $flowdockName
       description: $description
       skillsIds: $skillsIds
+      projectsChampionedIds: $projectsChampionedIds
     ) {
       id
       name
@@ -32,15 +34,23 @@ const updateUserQuery = gql`
         id
         name
       }
+      projectsChampioned {
+        id
+        name
+      }
     }
   }
 `;
 
 function prepUserForUpdate(updatedUser) {
   const skillsIds = updatedUser.skills.map(skill => skill.id);
+  const projectsChampionedIds = updatedUser.projectsChampioned.map(
+    project => project.id
+  );
   return {
     ...updatedUser,
-    skillsIds
+    skillsIds,
+    projectsChampionedIds
   };
 }
 
--- a/src/authenticated/member/MemberProfileEditPage.js
+++ b/src/authenticated/member/MemberProfileEditPage.js
@@ -57,6 +57,9 @@ function prepUserForUpdate(updatedUser) {
 export default compose(
   withLoggedInUser,
   graphql(updateUserQuery, {
+    options: {
+      refetchQueries: ['getProject']
+    },
     props: ({ mutate }) => ({
       onEdit: updatedUser => {
         return mutate({ variables: prepUserForUpdate(updatedUser) });
--- a/src/authenticated/member/MemberProfileEditPage.js
+++ b/src/authenticated/member/MemberProfileEditPage.js
@@ -57,14 +57,14 @@ function prepUserForUpdate(updatedUser) {
 export default compose(
   withLoggedInUser,
   graphql(updateUserQuery, {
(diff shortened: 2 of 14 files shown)


src/forms/AutocompleteChip.js
```diff
@@ -13,22 +14,26 @@ export default class AutocompleteChip extends React.Component {
...
+      })),
+      value: null,
+      onChange: this._handleSelectionChange,
+      onCreateOption: this._handleCreate,
+      placeholder: this.props.placeholder,
+      isSearchable: true,
+      menuPortalTarget: document.body
+    };
+
+    return this.props.onCreate ? (
```
wdoug, line 33: Nice

src/forms/EditableProjects.js
```diff
@@ -0,0 +1,65 @@
...
+const allProjectsQuery = gql`
+  query allProjects {
+    allProjects {
+      id
+      name
+    }
+  }
+`;
+
+const createProjectQuery = gql`
```
wdoug: Let's not allow the creation of projects through this dropdown. We specifically want to limit the amount of projects that are created

src/forms/EditableUsers.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+  return (
+    <Link style={{ textDecoration: 'none' }} to={`/volunteers/${item.id}`}>
+      {chip}
+    </Link>
+  );
+};
+
+export default compose(
+  graphql(allUsersQuery, {
```
wdoug, line 42: You can delete this `graphql` HOC

src/forms/EditableUsers.js
```diff
@@ -0,0 +1,53 @@
...
+    refetchQueries: ['allUsers'],
+    props: ({ mutate }) => ({
+      createChip: null,
+      ItemComponent: UserChip
+    })
+  }),
+  graphql(allUsersQuery, {
+    refetchQueries: ['allUsers'],
```
wdoug: I think that `refetchQueries` is only useful for mutations, right?

src/authenticated/member/MemberProfile.js
```diff
@@ -113,14 +114,14 @@ const MemberProfile = ({ user, onFormDataChange, editing }) => {
         <div className="Member-skills-left">
           <h2>Projects</h2>
```
wdoug: Let's change the title here to clarify that these are projects that the user is a champion of. We may want to have a separate field in the future for just project involvement. Do you think just `Projects Championed` is adequate?

cvanem: That works for me

cvanem:
@wdoug Please review the recent changes.  Note, this is the first time I have worked with GraphQL, so please double check the graphql updates.

src/authenticated/project/ProjectSection.js
```diff
@@ -46,7 +46,7 @@ const ProjectSection = ({ project, editing, onFormDataChange }) => {
         label="Add Champion"
         {...commonProps}
       />
-
+
```
wdoug: This is a weird change. I would have thought that prettier would have removed trailing whitespace

cvanem: Yes it should have.  It looks like I must have committed the changes without installing the modules first?  I fixed this, but I wonder if there is a way to prevent this from happening?

src/forms/EditableUsers.js
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <Link style={{ textDecoration: 'none' }} to={`/volunteers/${item.id}`}>
+      {chip}
+    </Link>
+  );
+};
+
+export default compose(
+  graphql(allUsersQuery, {
+    refetchQueries: ['allUsers'],
```
wdoug: We can delete this `refetchQueries` item here as I believe it only applies to mutations

src/authenticated/project/ProjectEditPage.js
```diff
@@ -48,7 +50,8 @@ const ProjectEditPage = compose(
         mutate({
```
wdoug, line 50: We want this `updateProjectQuery` graphql HOC to refetch `getUser` since those queries could now have stale data after this mutation happens i.e.

```js
graphql(updateProjectQuery, {
// ..
  options: {
    refetchQueries: ['getUser']
  }
})

File: src/authenticated/member/MemberProfileEditPage.js [hunk 9]
```diff
@@ -32,15 +34,23 @@ const updateUserQuery = gql`
...
+  const projectsChampionedIds = updatedUser.projectsChampioned.map(
+    project => project.id
+  );
   return {
     ...updatedUser,
-    skillsIds
+    skillsIds,
+    projectsChampionedIds
   };
 }
```
wdoug, line 56: Similarly, we want the subsequent `updateUserQuery` graphql HOC to refetch `getProject` since those queries could now have stale data after this mutation happens i.e.

```js
graphql(updateUserQuery, {
// ..
  options: {
    refetchQueries: ['getProject']
  }
})
```

I suppose another option would be to use Apollo's APIs to directly update the cached client-side data, but this is simpler and should be sufficiently performant

src/authenticated/project/ProjectEditPage.js
cvanem: Ok, does getUser need to be defined in the updateProjectQuery as well or is that handled somewhere else already?

src/forms/EditableProjects.js
```diff
@@ -0,0 +1,47 @@
...
+  );
+};
+
+export default compose(
+  graphql(allProjectsQuery, {
+    skip: props => !props.editing,
+    props: ({ data: { allProjects, loading } }) => ({
+      allOptions: allProjects,
+      allOptionsLoading: loading,
+      ItemComponent: ProjectChip
```
wdoug: Hmm. This is inconvenient. When `skip` resolves to `true` this props function doesn't get run which makes `ItemComponent` not get passed causing the default `ItemComponent` be used which means the styling and linking doesn't work

src/authenticated/project/ProjectEditPage.js
wdoug: Nope, it's the name of a graphql function (not sure what the actual terminology is). You can see an example of a query using it [here](https://github.com/codefordenver/members/blob/a9c1e136ccfca82257e7f452512eb95d46948635/src/authenticated/member/MemberProfilePage.js#L6). It can just be referenced by string and apollo will mark all the queries that request it as stale and refetch them if that data is used in a rendered component

src/forms/EditableProjects.js
wdoug: This would be a case that would be a bit easier to handle with the apollo [Query component](https://www.apollographql.com/docs/react/essentials/queries.html#basic). Another option would be to have a little intermediary component that passes in the `ItemComponent` prop. If we don't care about displayNames we could even inline `props => <EditableList {...props} ItemComponent={ProjectChip}>` on line 47.

src/forms/EditableUsers.js
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <Link style={{ textDecoration: 'none' }} to={`/volunteers/${item.id}`}>
+      {chip}
+    </Link>
+  );
+};
+
+export default compose(
+  graphql(allUsersQuery, {
+    skip: props => !props.editing,
```
wdoug, line 43: We have the same situation here with the `ItemComponent`

wdoug:
Styling seems good

src/authenticated/project/ProjectEditPage.js
```diff
@@ -43,12 +45,16 @@ const ProjectEditPage = compose(
     props: ({ data: { Project } }) => ({ project: Project })
   }),
   graphql(updateProjectQuery, {
+    options: {
+      refetchQueries: ['getUser']
```
wdoug: Oh whoops, I am realizing `getUser` is the name of the query, not the type. I'm not sure if we want [`User` or `user`](https://github.com/codefordenver/members/blob/a9c1e136ccfca82257e7f452512eb95d46948635/src/authenticated/member/MemberProfilePage.js#L7). You may want to test both and see which one works

wdoug: Nevermind. It seems to be working like this.

src/authenticated/member/MemberProfileEditPage.js
wdoug, line 56: We still want `refetchQueries: ['getProject']` on the `updateUserQuery` graphql query HOC. I think that is the last thing to update and then we can get this merged

cvanem: Ok, I thought I added that before but I guess not.  This PR was a little messy. It should be ready to merge now.

cvanem: Hmm, the project doesn't seem to be updating...

wdoug (approved):
This looks great! I'm going to go ahead and merge this so I can integrate the changes from this branch with the Typescript branch

PR #278, opened 2018-11-27
WIP: Auth Refactoring

dchao19:
This PR closes a whole mess of auth related issues.

#### What does this PR do?
 - Refactors the authentication system in a few crucial ways:
     * Entirely removes the `withAuthSession` and `withLoggedInUser` HOCs that have been haunting our component tree for months
     * Introduces another route `/callback` to handle the authentication result from Auth0, instead of handling it in the `LoginAuth0` component included in the header. This is the recommended approach. 
     * Adds a new `AuthenticationContext` which can provide child component will access to the userProfile and other authentication keys as needed
     * Refactors the Header, Login, MyProfilePage, MemberProfileEditPage components to use `AuthenticationContext` instead of the auth HOCs
     * Starts using the `Query` and `Mutation` component wherever feasible in the Auth system (notably, not in the AuthCallback component. This is because the queries / mutations need to happen in a specific order and the component model doesn't support this)
      
##### Things that still need to be worked out
  - The `MyProfilePage` and `MemberProfileEditPage` component are disgustingly complicated. The `MyProfilePage` gets its data from the AuthenticationContext, and thus the LocalStorage, which makes editing the userProfile more complicated. Any updates to the profile must update the server, the AuthenticationContext, and the LocalStorage.  
     * For instance, the MemberProfileEditPage now has to update the AuthContext and the local storage in the "onCompleted" prop. To fix this, I introduced a local storage side effect in Context, but I don't think this is a good approach. 
  - To effectively use the AuthenticationContext with an already saved user-session, in the `App` main component `componentDidMount()` I hydrate the context with the LocalStorage data, which makes me wonder if we even need to use Context at all, or if it just makes things unnecessarily complicated.
 - I was having trouble with getting different "Sign Up" and "Login" buttons to work, but could be my specific Auth0 configuration
 - Because of these issues, I don't think this PR is ready to merge. I would like more feedback before proceeding.


dchao19:
A crucial next step is to test this in a broader staging environment like the deploy preview. This will require @wdoug or another CfD organizer to add "/callback" to the list of allowed callback URLS.

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
 });
 
 class App extends Component {
+  constructor(props) {
```
wdoug: Let's put all of this logic in its own component, for example a `AuthenticationProvider` component that just renders its children

src/authenticated/member/MemberProfileEditPage.js
```diff
@@ -44,18 +45,41 @@ function prepUserForUpdate(updatedUser) {
...
-)(MemberProfile);
+class MemberProfileEditPage extends Component {
+  render() {
+    return (
+      <AuthenticationContext.Consumer>
+        {context => (
+          <Mutation
+            mutation={updateUserQuery}
+            onCompleted={({ updateUser }) => {
+              context.setCurrentUserProfile(updateUser);
```
wdoug: How does this get on here? I don't see `setCurrentUserProfile` in the context provider value

src/authenticated/member/MemberProfileEditPage.js
```diff
@@ -44,18 +45,41 @@ function prepUserForUpdate(updatedUser) {
...
+            }}
+          >
+            {mutate => {
+              const onEdit = updatedUser =>
+                mutate({
+                  variables: {
+                    ...prepUserForUpdate(updatedUser),
+                    id: context.authData.userId
```
wdoug: Why set the `id` here?

src/authenticated/member/MemberProfileEditPage.js
```diff
@@ -44,18 +45,41 @@ function prepUserForUpdate(updatedUser) {
...
-      }
-    })
-  }),
-  withEditPage({
-    renameProps: {
-      formData: 'user'
-    }
-  })
-)(MemberProfile);
+class MemberProfileEditPage extends Component {
```
wdoug: If this component is specific to users editing themselves should we make the component name more specific? We have a `MyProfilePage` component. Should we make this `MyProfileEditPage`? Also, we should refactor all these different versions of these similar pages and clean them up

wdoug: The refactor can happen in another PR

src/authenticated/member/MemberProfileEditPage.js
wdoug: I'm not a fan of having this `onCompleted` logic here. Won't that only set the user after it is mutated? We want the user data to be present here without it. We should have a `Query` component wrapper that passes it in.

wdoug: The `Query` component could be part of the `AuthenticationProvider` component. Not sure if that would be a bad idea or not...

src/authenticated/member/MyProfilePage.js
```diff
@@ -20,22 +21,26 @@ const userQuery = gql`
...
-  withViewPage({
-    renameProps: {
-      formData: 'user'
-    }
-  })
-)(MemberProfile);
+const MyProfilePage = () => (
+  <AuthenticationContext.Consumer>
+    {context => {
+      console.log(context);
```
wdoug: We can remove the console log

src/constants/storageKeys.js
```diff
@@ -1,4 +1,5 @@
-export const ACCESS_TOKEN_KEY = 'access_token';
```
wdoug: Why are these lines considered changed? I hope it's not because of line endings...

wdoug:
I just had a thought that if we switched to Prisma we could probably add a custom graphql type that is the `CurrentUser` that returns the authenticated user. I don't know what implementing that would be like, but I wonder if it would simplify some of our current user logic.

src/unauthenticated/LoggedOutRoutes.js
```diff
@@ -2,10 +2,13 @@ import React from 'react';
 import { Route, Switch } from 'react-router-dom';
 import LandingPage from './LandingPage';
 import NoMatchPage from '../shared-components/NoMatchPage';
+import AuthCallBack from '../utils/authentication/AuthCallback';
+import AuthService from '../utils/authentication/authService';
 
 const LoggedOutRoutes = () => (
   <Switch>
     <Route exact path="/" component={LandingPage} />;
+    <Route exact path="/callback" component={AuthCallBack} />
```
wdoug: Can we add this config to the auth0 setup that is run as part of the initial bootstrapping?

src/utils/authentication/AuthCallback.js
```diff
@@ -0,0 +1,105 @@
...
+
+  async componentDidMount() {
+    try {
+      // Extract the id_token and access_token from the Auth0 federation
+      // Internally, the Auth0 library search does another call to get the user profile info
+      // so "parseAuthenticationResult" is async
+      const authResult = await AuthService.parseAuthenticationResult();
+
+      // Send a request to the backend to get the token used to authenticate with Graphcool
+      // and create the user in the database if necessary
```
wdoug, line 38: Nice comments in here

src/utils/authentication/AuthCallback.js
```diff
@@ -0,0 +1,105 @@
...
+      const authData = {
+        auth0AccessToken: authResult.accessToken,
+        graphcoolToken: token,
+        userId: id,
+        expiresAt: JSON.stringify(authResult.expiresIn * 1000 + Date.now()),
+        userProfile: user
+      };
+      this.context.setCurrentUser(authData);
+
+      this.props.history.replace('/');
```
wdoug, line 66: Do you know if there is a simple way to get the location before the authentication process was started? That could be a follow up card

src/utils/authentication/AuthCallback.js
```diff
@@ -0,0 +1,105 @@
...
+    } catch (e) {
+      console.error(e);
```
wdoug: I think we are missing two different user error notifications from the previous logic here. We're missing the standard error alert and the alert that a user already exists with that email ([seen here](https://github.com/codefordenver/members/pull/278/files?utf8=%E2%9C%93&diff=unified#diff-03247827a8b105df8686de5830445287L85))

src/utils/authentication/authService.js
```diff
@@ -0,0 +1,98 @@
...
+  signUp = () => {
+    this.webAuth.authorize();
+  };
+
+  parseAuthenticationResult = () => {
+    return new Promise((resolve, reject) => {
+      this.webAuth.parseHash((err, authResult) => {
+        // Hash is in window.location.hash
+        if (err) {
+          reject(err);
```
wdoug: Nice

src/utils/authentication/authService.js
```diff
@@ -0,0 +1,98 @@
...
+  BEARER_TOKEN,
+  EXPIRES_AT_KEY,
+  USER_ID,
+  USER_PROFILE
+} from '../../constants/storageKeys';
+
+const env = getEnvironmentVariables();
+
+class AuthService {
+  returnUrl = window.location.origin;
```
wdoug: Is this used anywhere?

src/utils/authentication/authService.js
```diff
@@ -0,0 +1,98 @@
...
+      auth0AccessToken: localStorage.getItem(ACCESS_TOKEN_KEY),
+      graphcoolToken: localStorage.getItem(BEARER_TOKEN),
+      userId: localStorage.getItem(USER_ID),
+      expiresAt: localStorage.getItem(EXPIRES_AT_KEY),
+      userProfile
+    };
+  };
+}
+
+export default new AuthService();
```
wdoug: What's the purpose of the class singleton here? It seems like these are all independent methods that could just be exported as individual functions. What do you think?

src/utils/authentication/authService.js
```diff
@@ -0,0 +1,98 @@
...
+    responseType: 'token id_token',
+    scope: 'openid email profile'
+  });
+
+  login = () => {
+    this.webAuth.authorize();
+  };
+
+  signUp = () => {
+    this.webAuth.authorize();
```
wdoug: I think we need to pass `{ initialScreen: 'signUp' }` to get our customized auth0 login page to show the signup tab when navigated to.

src/authenticated/member/MemberProfileEditPage.js
wdoug: Oh, nevermind on the question of how it gets there. I see it now

src/utils/authentication/authContext.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const AuthenticationContext = React.createContext({
```
wdoug: The data shape here is inconsistent with the value used in the App: 
```js
{
        authData: {},
        isAuthenticated: this.isAuthenticated,
        setCurrentUser: this.setCurrentUser,
        setCurrentUserProfile: this.setCurrentUserProfile
}
```

src/App.js
wdoug: Maybe `AuthProvider` is nicer. Shorter at least...

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
...
+    super(props);
+
+    this.state = {
+      authContext: {
```
wdoug: If we pull this out into a separate `AuthProvider` component we can flatten this state and remove the nested authContext object, which makes updates simpler

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
...
+        isAuthenticated: this.isAuthenticated,
+        setCurrentUser: this.setCurrentUser,
+        setCurrentUserProfile: this.setCurrentUserProfile
+      }
+    };
+  }
+
+  componentDidMount = () => {
+    // Sync auth session stored in the localStorage with Context
+    // TBH, I'm not sure if we even need context
```
wdoug: We need a way to notify components that the value has changed. That's what we're using context for

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
...
+    this.setCurrentUser(AuthService.getSessionAuthData());
```
wdoug: Instead of calling this here and having a rerender, can we just call `AuthService.getSessionAuthData()` in the constructor?

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
...
+
+  isAuthenticated = () => {
+    const {
+      userId,
+      expiresAt,
+      auth0AccessToken,
+      graphcoolToken
+    } = this.state.authContext.authData;
+    return (
+      Date.now() < expiresAt && userId && auth0AccessToken && graphcoolToken
```
wdoug: Any reason why we wouldn't put this logic in the `AuthService`?

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
...
+      authContext: {
+        ...this.state.authContext,
+        authData
+      }
+    });
+  };
+
+  setCurrentUserProfile = userProfile => {
+    AuthService.setAuthProfileSession(userProfile); // I don't really like this side-effect being here
+    this.setState({
```
wdoug: These calls to `setState` should probably be using the function argument [updater format](https://reactjs.org/docs/react-component.html#setstate) as the state update is based on current state:
```js
this.setState(state => ({
    authContext: {
        state.authContext,
        authData: {
          state.authContext.authData,
          userProfile
        }
    }
}));
```

src/header/Header.js
```diff
@@ -1,46 +1,70 @@
...
+              {context.isAuthenticated() ? (
+                <React.Fragment>
+                  <Grid item>
+                    <Link className="Header-link" to="/volunteers">
+                      All Users
+                    </Link>
+                    <Link className="Header-link" to="/projects">
+                      All Projects
+                    </Link>
+                    {userIsAdmin(context.authData.role) && (
```
wdoug: We could embed the `isAdmin` function on the context as well. That would probably be nice

src/header/Header.js
```diff
@@ -1,46 +1,70 @@
...
+                      <Link className="Header-link" to="/admin">
+                        Admin Resources
+                      </Link>
+                    )}
+                  </Grid>
+                  <MenuList
```
wdoug: Should this be wrapped in a `<Grid item>` as well?

src/header/Header.js
```diff
@@ -1,46 +1,70 @@
...
+                    avatar={context.authData.userProfile.picture}
+                    username={context.authData.userProfile.name}
+                    logout={() => AuthService.logout()}
+                  />
+                </React.Fragment>
+              ) : (
+                <Grid item>
+                  <Button color="secondary" onClick={() => AuthService.login()}>
```
wdoug: No need for the arrow function, just use `onClick={AuthService.login}`. Man, this is so much nicer!

src/header/Header.js
```diff
@@ -1,46 +1,70 @@
...
+                    Login
+                  </Button>
+                  <Button
+                    color="secondary"
+                    onClick={() => AuthService.signUp()}
```
wdoug: Same as above, no need for the arrow function

src/utils/authentication/AuthCallback.js
```diff
@@ -0,0 +1,105 @@
...
+        name
+      }
+    }
+  }
+`;
+
+class AuthCallback extends Component {
+  static contextType = AuthenticationContext;
+
+  async componentDidMount() {
```
wdoug, line 30: Nice work here. This is so much nicer as an async function!

src/utils/authentication/authService.js
```diff
@@ -0,0 +1,98 @@
...
+    localStorage.removeItem(BEARER_TOKEN);
+    localStorage.removeItem(USER_PROFILE);
+  };
+
+  getBearerToken = () => {
+    localStorage.getItem(BEARER_TOKEN);
+  };
+
+  setAuthProfileSession = user => {
+    localStorage.setItem(USER_PROFILE, JSON.stringify(user));
```
wdoug: I think I'd prefer to just query for this data every time instead of storing it in local storage. I think that will simplify a few things and make it so we don't have to worry about whether a user updated their profile in another browser or on another device.

src/header/Header.js
```diff
@@ -1,46 +1,70 @@
...
-import { compose } from 'react-apollo';
-import Login from './Login';
 import AppBar from '@material-ui/core/AppBar';
 import Toolbar from '@material-ui/core/Toolbar';
+import Button from '@material-ui/core/Button';
 import Grid from '@material-ui/core/Grid';
 import logo from '../images/cfd-circle-icon-white.png';
 import userIsAdmin from '../utils/userIsAdmin';
-import withLoggedInUser from '../utils/withLoggedInUser';
-import withAuthSession from '../utils/withAuthSession';
```
wdoug, line 11: Let's delete the `withLoggedInUser` and `withAuthSession` files

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
...
+      expiresAt,
+      auth0AccessToken,
+      graphcoolToken
+    } = this.state.authContext.authData;
+    return (
+      Date.now() < expiresAt && userId && auth0AccessToken && graphcoolToken
+    );
+  };
+
+  setCurrentUser = authData => {
```
wdoug: These methods are confusingly named. Should this be called `setAuthData` or `setAuthMetadata` instead?

src/authenticated/member/MemberProfileEditPage.js
wdoug: Alternatively, instead of putting the `Query` component in the `AuthProvider`, we might want to just have different `Query`s for individual data needs for the components that need parts of the logged in user.

src/unauthenticated/LoggedOutRoutes.js
wdoug: Also, can we make this a bit more descriptive than just callback? Maybe `/auth-callback` or `/login-callback` or something?

src/utils/authentication/authService.js
```diff
@@ -0,0 +1,98 @@
...
+    localStorage.setItem(EXPIRES_AT_KEY, expiresAt);
+    localStorage.setItem(BEARER_TOKEN, graphcoolAccessToken);
+  };
+
+  logout = () => {
+    localStorage.removeItem(ACCESS_TOKEN_KEY);
+    localStorage.removeItem(USER_ID);
+    localStorage.removeItem(EXPIRES_AT_KEY);
+    localStorage.removeItem(BEARER_TOKEN);
+    localStorage.removeItem(USER_PROFILE);
```
wdoug: I think we want to also make sure we're navigating away and resetting the app state when we log out. We might want to just force a hard refresh at this point to make sure it is all cleared out. Maybe something like just `window.location.href = '/'`

src/App.js
```diff
@@ -35,21 +37,76 @@ const styles = theme => ({
...
+      authContext: {
+        ...this.state.authContext,
+        authData: {
+          ...this.state.authContext.authData,
+          userProfile
+        }
+      }
+    });
+  };
+
```
wdoug: Let's also add a listener to `storage` changes somewhere (probably in the `AuthProvider`) to update the state/handle logouts if the storage changes in another tab. See current code for this [here](https://github.com/codefordenver/members/blob/ee4580e3a66b83c2ced90cca0ce95433111b6b4f/src/utils/withAuthSession.js#L69-L74)

wdoug:
@dchao19 this is looking great! Sooooo much better than the crazy hacks that we have currently!!!

Regarding your "Things that still need to be worked out":

> - The `MyProfilePage` and `MemberProfileEditPage` component are disgustingly complicated. The `MyProfilePage` gets its data from the AuthenticationContext, and thus the LocalStorage, which makes editing the userProfile more complicated. Any updates to the profile must update the server, the AuthenticationContext, and the LocalStorage.  
>      - For instance, the MemberProfileEditPage now has to update the AuthContext and the local storage in the "onCompleted" prop. To fix this, I introduced a local storage side effect in Context, but I don't think this is a good approach. 

I'd like to request the user data on page load instead of storing it in local storage. After looking at it again, I think we should actually just request the user data for the few places it's needed with standard graphql co-located queries. I'd like to change the `MyProfilePage` to just wrap the `MemberProfilePage` and pass in the id from Auth as the variable for the query that component is already making.

I'd like to refactor the form handling for the site and the combine the `withEditPage`, `withCreatePage`, and `withViewPage` into a single component of some form, which I think would further be able to greatly simplify things on these pages.

>  - To effectively use the AuthenticationContext with an already saved user-session, in the `App` main component `componentDidMount()` I hydrate the context with the LocalStorage data, which makes me wonder if we even need to use Context at all, or if it just makes things unnecessarily complicated.

I don't think we need it, but I think we'll still want it. Here's the way I see it: 

We need to have the app rerender when the auth data changes so that it renders the correct thing (we had bugs around this in the past which is why we had that whole crazy auth HOC in the first place). As far as I can tell this requires us to store data in state up near the top of the application above everything that needs auth info. We could then either:
1. Pass this data and callbacks to update it down through props
2. Pass down callbacks to update the data down through props and just grab the data off localStorage as needed
3. Put the data and update callbacks on context so any child can utilize it

Option 1 would work fine, but would probably get a little annoying plumbing that data through several components.

Option 2 is basically referencing global data and has a weird side effect setup that is harder to test.

Option 3 might actually be the easiest to set up test abstractions for and also allows us to skip passing the props through potentially several layers of components which is nice.

> - I was having trouble with getting different "Sign Up" and "Login" buttons to work, but could be my specific Auth0 configuration

This functionality requires a custom Auth0 log in page (annoyingly). In the [readme](https://github.com/codefordenver/members/blob/master/README.md) we have documented:

(To enable Log In/ Sign Up functionality, please go to your Auth0 dashboard, navigate to "Hosted Pages", and enter the following line within the 'lock' variable.

        initialScreen: config.extraParams.initial_screen,
)

> - Because of these issues, I don't think this PR is ready to merge. I would like more feedback before proceeding. 

Well, I just wrote up a ton of feedback. 😉 Let's coordinate to get this merged

src/App.js
wdoug: I have now provided a more detailed response to this question in [this comment](https://github.com/codefordenver/members/pull/278#issuecomment-444757012)

wdoug:
We might want to eventually use Apollo Client's apollo-link-state to [store client-side data](https://www.apollographql.com/docs/react/essentials/local-state.html) (which I believe is actually being merged into the core of apollo going forward). See [apollo's new full stack tutorial section on local state management](https://www.apollographql.com/docs/tutorial/local-state.html) ([code reference](https://github.com/apollographql/fullstack-tutorial/blob/master/final/client/src/resolvers.js)) for an example of adding a [`isLoggedIn` client extension of the graphql api](https://github.com/apollographql/fullstack-tutorial/blob/master/final/client/src/resolvers.js)

dchao19:
Pretty much the only reason why I decided to put the user profile in local storage instead of using immediate queries in components that need the data, is that I really didn't want the `Header` component to have to query the user profile just for the profile picture and if the user is an admin or user. In my mind, it would be better to store this data in Auth0's custom metadata so that the role/profile picture can be queried when requesting the ID Token instead of relying on the backend, and I was trying to decide between implementing the query in the header component or committing the profile to local storage.

wdoug:
Let's update the docs to point out that the netlify callback needs to be updated for developers that have worked on this in the past

src/header/Header.js
```diff
@@ -1,46 +1,92 @@
...
+                          avatar={data.user.picture}
+                          username={data.user.name}
+                          logout={AuthService.logout}
+                        />
+                      </React.Fragment>
+                    );
+                  }}
+                </Query>
+              ) : (
+                <Grid item>
```
wdoug: We probably don't want to show the login and sign up buttons if we are on the callback page actively being logged in. In that case would we just show a spinner in the Header?

wdoug (approved):
Looks great. Lets merge this bad boy

PR #287, opened 2018-12-27
Fix defects while creating projects

### LINKED ISSUE #285: Creating new projects doesn't support adding champions right away
Projects can be created at `/projects/create`

wdoug:
This PR closes #285 

Depends on #286 

#### What does this PR do?

- Fixes a few defects while creating a new project
  - Makes adding champions to new projects actually work
  - Makes skills and champions optional for new projects
- Converts the project pages to typescript
- Adds some tests to the create project page 
- Makes a few new test helpers


src/authenticated/project/ProjectCreatePage.tsx
```diff
@@ -0,0 +1,46 @@
...
+import {
+  CreateProjectHOC,
+  ProjectSectionFieldsFragment
+} from '../../generated-models';
+
+const ProjectCreatePage = compose(
+  CreateProjectHOC({
+    props: ({ mutate }) => ({
+      onCreate: (newProject: ProjectSectionFieldsFragment) => {
+        if (!newProject.repoName) {
```
dchao19, line 13: I feel like at some point this form of validation should be moved into the actual form - not on the create event, but a problem for a later time I suppose.

dchao19 (approved):

src/authenticated/project/ProjectCreatePage.tsx
wdoug: Haha good point

PR #294, opened 2019-02-05
Hide "Log In" / "Sign Up" buttons during the callback

### LINKED ISSUE #293: Hide Login and Signup buttons during the auth callback
This issue is a regression from previous behavior. 

Behavior before the auth refactor:
 - The user clicks the "Login" or "Signup" button
 - The user is returned to the original page and the "Login" or "Signup" buttons are grayed out with a circular progress indicator adjacent to them
 - The authentication process is complete, the user is redirected, and the "Login" / "Signup" buttons are replaced with navigation links

Behavior after the auth refactor
 - The user clicks the "Login" or "Signup" button
 - The user is redirected to an intermediate "callback" page where the auth result is handled, but the login and signup buttons are still visible.

An accepted resolution for this issue should either hide / disable / gray out the login or signup buttons during the callback phase.

dchao19:
This PR doesn't closes #293

#### What does this PR do?
This PR hides the "Log In" and "Sign Up" buttons during the callback by adding a `isLoggingIn` and a respective `setLoggingIn` field and method to the context. It is called by the AuthCallback to hide the "Log In" and "Sign Up" buttons. 
This PR updates the "Header" component to use Typescript and separate it into slightly smaller components. 
This PR updates the "AuthProvider" component to use Typescript
This PR tweaks the null / undefined behavior of the context and the AuthProvider.

### DIFF
--- a/src/header/Header.tsx
+++ b/src/header/Header.tsx
@@ -22,7 +22,7 @@ const AuthButtons = ({ isLoggingIn = false }) => {
         Log In
       </Button>
       <Button color="secondary" onClick={AuthService.signUp}>
-        Signup
+        Sign Up
       </Button>
     </Grid>
   );
--- a/src/header/Header.tsx
+++ b/src/header/Header.tsx
@@ -19,12 +19,10 @@ const AuthButtons = ({ isLoggingIn = false }) => {
   return (
     <Grid item>
       <Button color="secondary" onClick={AuthService.login}>
-        {' '}
-        Log In{' '}
+        Log In
       </Button>
       <Button color="secondary" onClick={AuthService.signUp}>
-        {' '}
-        Signup{' '}
+        Signup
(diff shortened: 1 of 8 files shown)


src/header/Header.tsx
```diff
@@ -19,12 +19,10 @@ const AuthButtons = ({ isLoggingIn = false }) => {
...
     <Grid item>
       <Button color="secondary" onClick={AuthService.login}>
-        {' '}
-        Log In{' '}
+        Log In
       </Button>
       <Button color="secondary" onClick={AuthService.signUp}>
-        {' '}
-        Signup{' '}
+        Signup
```
wdoug: Can we be consistent with our spacing and say `Sign Up`?

wdoug (approved):
LGTM

PR #298, opened 2019-02-12
Support non codefordenver org repos

### LINKED ISSUE #280: Allow non-Code for Denver repos as valid Github Repos
Current idea for implementation is to check the Github link provided, and if it is a full link to a github (instead of just a Repo name), use that link. Otherwise, prepend with https://github.com/codefordenver

@wdoug - for non cfd github repos, will they also need to have a separate waffle.io link?

wdoug:
This PR closes #280 

#### What does this PR do?

Adds support for linking to repos that are under different orgs/users (water beacon is the example this is targetting: https://github.com/mosesmc52/waterbeacon)


wdoug:
It sounds like we want to support non-waffle tools

src/forms/EditableLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+  name,
+  label,
+  linkName,
+  linkTarget,
+  editing,
+  onChange = () => {}
+}) => {
+  if (!editing) {
+    return value ? (
+      <div>
```
wdoug: Let's remove the div

wdoug:
Alright, I'm going to go ahead and merge this. It should be ready now, and we could definitely use some of this new functionality for the move off waffle.io

PR #299, opened 2019-02-12
Add initial member since info to member profile

### LINKED ISSUE #34: Add member since ...
As a member I would like to see how long I've been contributing to CFD. I would I also like a way to edit my contribution beginning time to reflect actual time.

wdoug:
This PR closes #34 

#### What does this PR do?

Adds a member since section to the member profile page

This would have been useful recently when I was trying to track down which user was which between two people that had the same name.

### DIFF
--- a/src/authenticated/member/Member.css
+++ b/src/authenticated/member/Member.css
@@ -28,6 +28,11 @@
   margin: 20px;
 }
 
+.Member-since {
+  color: #999;
+  font-size: 0.8rem;
+}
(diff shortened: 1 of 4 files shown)


wdoug:
@dviramontes I know this isn't quite what was originally described in #34, but how does this sound for now?

dviramontes (approved):
lets do it ✅

dchao19 (approved):

PR #306, opened 2019-04-16
Padding fix on Header & Paragraph

anotherpatrickcollins:
Fixes the 0 padding on the header and paragraph tags

### DIFF
--- a/src/unauthenticated/LandingPage.css
+++ b/src/unauthenticated/LandingPage.css
@@ -3,5 +3,5 @@
 }
 
 .leftPad {
-  margin-eft: 30px;
+  margin-left: 30px;
 }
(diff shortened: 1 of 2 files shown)


src/unauthenticated/LandingPage.css
```diff
@@ -1,3 +1,7 @@
 .LandingPage-members-pic {
   max-width: 100%;
 }
+
+.leftPad {
```
wdoug, line 5: Nice name

src/unauthenticated/LandingPage.css
```diff
@@ -1,3 +1,7 @@
...
+  margin-eft: 30px;
```
wdoug:
```suggestion
  margin-left: 30px;
```

wdoug (approved):
:shipit:

PR #322, opened 2019-05-17
Delete github issues

### LINKED ISSUE #318: Delete code pulling in `ready` issues from github since that won't work properly without using waffle.io

wdoug:
Closes #318

#### What does this PR do?

Deletes the pulled in github issues with "ready" tags since waffle shut down and this is no longer relevant. 😢


anotherpatrickcollins (approved):
Cool codecov red thing. Not sure if I need to be concerned.
hack the planet.

PR #327, opened 2019-07-09
All hooks all the time

wdoug:
### What does this PR do? 

Uses React hooks for just about everything


package.json
```diff
@@ -30,9 +30,10 @@
     "graphql": "^14.0.2",
     "graphql-tag": "^2.10.0",
     "marked": "^0.5.1",
-    "react": "^16.8.2",
+    "react": "^16.9.0-alpha.0",
```
wdoug, line 33: This is for avoiding a ton of warnings about needing to use `act` https://github.com/testing-library/react-testing-library/issues/281

wdoug:
The new version of react-apollo also just added hooks, but they don't seem to support suspend yet. See: https://github.com/apollographql/react-apollo/pull/2892

src/authenticated/LoggedInRoutes.tsx
```diff
@@ -16,52 +16,49 @@ import ProjectPage from './project/ProjectPage';
...
+            <Route exact path="/me" component={MyProfilePage} />
+            <Route exact path="/me/edit" component={MemberProfileEditPage} />
+            <Route exact path="/flowdock" component={FlowdockLanding} />
+            <Route exact path="/volunteers" component={UsersListPage} />
+            <Route exact path="/volunteers/:id" component={MemberProfilePage} />
+            <Route path="/projects/create" component={ProjectCreatePage} />
+            <Route path="/projects/:id/edit" component={ProjectEditPage} />
+            <Route path="/projects/:id" component={ProjectPage} />
+            <Route path="/projects" component={ProjectsListPage} />
+            {/* <Route path="/projects" component={GeneralProjectPage} /> */}
```
MikeLuDev: Commented code could be removed

src/authenticated/onboarding/OnboardingPage.tsx
```diff
@@ -25,56 +28,47 @@ function formatMemberForMutation(updatedUser: MemberProfileFragmentFragment) {
...
+  const updateMemberMutation = useMutation<UpdateUserMutation>(
+    UpdateUserDocument,
+    {
+      refetchQueries: ['editableUsersList']
+    }
+  );
+  const { data } = useCustomQuery<GetUserQuery>(GetUserDocument, {
+    variables: { id: authContext.authData.userId }
+  });
+  if (!data || !data.user) return null;
```
MikeLuDev, line 42: I see this pattern a lot in your code. It is probably always better to inform the user that there was no data found rather than returning `null`.

src/authenticated/onboarding/OnboardingPage.tsx
```diff
@@ -25,56 +28,47 @@ function formatMemberForMutation(updatedUser: MemberProfileFragmentFragment) {
...
-                    />
-                  </Form>
-                )}
-              />
-            );
-          }}
-        </GetUserComponent>
+    <Formik
+      initialValues={data.user}
+      onSubmit={async (updatedMember, actions) => {
```
MikeLuDev, line 47: I've seen this a few times too. It would be cleaner to handle the logic outside of `onSubmit` in a separate function, `handleSubmit` for instance. I consider this a non-blocker however.

src/authenticated/project/ProjectsListPage.tsx
```diff
@@ -2,20 +2,23 @@ import React from 'react';
...
   loading: boolean;
 }
 
-const ProjectsList: React.SFC<ProjectsListProps> = ({ projects, loading }) => {
-  if (loading) {
-    return <LoadingIndicator />;
-  }
+const ProjectsListPage: React.FC<ProjectsListProps> = () => {
+  const { data } = useCustomQuery<ProjectCardsQuery>(ProjectCardsDocument);
+  if (!data || !data.allProjects) return <LoadingIndicator />;
```
MikeLuDev: This doesn't follow the same pattern as the other cases of `useCustomQuery`. In the others you return `null`, but here a loading indicator is returned. As a user, this is more confusing than being given an empty screen because the loading indicator implies that the data is still coming.

src/authenticated/skill/SkillPage.tsx
```diff
@@ -9,16 +11,14 @@ type SkillPageProps = {
...
-    refetch(); // Refresh on page render so the data isn't stale
-    return { skill, loading };
-  }
-})(SkillDetails);
+const SkillPage: React.FC<SkillPageProps> = ({ match }) => {
+  const { data, refetch } = useCustomQuery<SkillPageQuery>(SkillPageDocument, {
+    variables: { id: match.params.id }
+  });
+  if (!data || !data.skill) return null;
+  refetch(); // Refresh on page render so the data isn't stale
```
MikeLuDev, line 19: :+1:

src/forms/EditableStatus.tsx
```diff
@@ -39,20 +39,20 @@ const EditableStatus: React.SFC<Props> = ({
...
   };
+  const selectValue = options.find(option => option.value === value);
   const selectProps = {
     'aria-label': 'status',
     className: 'EditableStatus',
     options: options,
-    value: options.find(option => option.value === value),
-    label: options.find(option => option.value === value),
+    value: selectValue,
+    label: selectValue,
```
MikeLuDev, line 48: :+1:

MikeLuDev (changes requested):
Some quick UX changes that should be addressed.

src/authenticated/onboarding/OnboardingPage.tsx
wdoug: This actually should never happen. This is just to appease the TypeScript type checker because I don't know enough advanced TypeScript to specify that the output at this point should never be undefined.

src/authenticated/project/ProjectsListPage.tsx
wdoug: Good catch.

src/authenticated/onboarding/OnboardingPage.tsx
wdoug: I think there might have been some typescript annoyance where the types for the function needed to be explicitly defined if pulled out of this inline context. Would be worth investigating again at some point

MikeLuDev (approved):

Issue #326, opened 2019-07-09
Filter projects based on status


wdoug:
- Add selection dropdown to filter projects based on status
- By default, do not show inactive projects

wdoug:
@anotherpatrickcollins how did we want to do the UX for this again?

PR #329, opened 2019-07-15
Graphql refactor

wdoug:
### What does this PR do? 
- Refactors client graphql to be defined in typescript directly instead of having the confusion of defining it in .graphql files and people not understanding how it worked
- Updates the graphql-code-generator library to take advantage of the latest and greatest
- Creates a few hooks for shared graphql queries and mutations
- Fixes some graphql test warnings about missing __typename


wdoug:
TODO
- [x] Go through comment TODOs
- [x] Update readme
- [x] Check up on new test warning - I'll look more into this separately

src/utils/commonGraphql.ts
```diff
@@ -0,0 +1,260 @@
...
+      // TODO: These didn't exist before, check if they are actually needed
+      refetchQueries: ['projectCards', 'projectsDrawer', 'editableUsersList']
+    }
+  );
+
+  return (newProject: ProjectCommonFragment) =>
+    createProject({
+      variables: formatNewProjectForCreate(newProject)
+    });
+}
```
MikeLuDev, line 254: Cool, one concern I had was that some queries might be reused between components, and that eliminating the old pattern would discourage DRY code. This file addresses that 👍

src/authenticated/DrawerContent.tsx
```diff
@@ -27,11 +25,21 @@ const styles = (theme: Theme) =>
     }
   });
 
+const DRAWER_PROJECTS = gql`
+  query projectsDrawer {
+    allProjects(orderBy: name_ASC) {
+      # TODO: Only show active projects
```
MikeLuDev, line 31: Comments like this would be better left out of the code and instead made into an issue on Github.

MikeLuDev (approved):
Looks great overall. Seems more in line with common patterns to keep queries in the same file as the single component which uses them.

Comments in code are not a blocker.

src/authenticated/DrawerContent.tsx
MikeLuDev: Added the issue outlined in this comment as a requirement in https://github.com/codefordenver/members/issues/326

src/utils/commonGraphql.ts
wdoug: Yeah there are different tradeoffs. In Relay.js for example, I believe the recommendation is to define every graphql query (and possibly mutations too) uniquely for the specific component that uses it. I think this setup is better for our use case currently though

PR #330, opened 2019-07-15
Convert remaining components to functional components

wdoug:
### What does this PR do? 

Converts the remaining class components that could be converted to functional components to functional components

### DIFF
--- a/src/shared-components/MarkdownFileRenderer.tsx
+++ b/src/shared-components/MarkdownFileRenderer.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import marked from 'marked';
 import LoadingIndicator from './LoadingIndicator';
 import './MarkdownFileRenderer.css';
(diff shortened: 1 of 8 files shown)


src/shared-components/MarkdownFileRenderer.tsx
```diff
@@ -17,56 +17,39 @@ async function getMarkdownHTML(fileName: string) {
...
+  getMarkdownHTML(fileName).then(html => setHtml(html));
+
+  return (
+    <div className="MarkdownFileRenderer">
+      {html ? (
+        [
+          <a
+            key="markdown-edit-link"
+            className="MarkdownFileRenderer-edit"
+            href={`https://github.com/codefordenver/members/edit/master/public/markdown/${fileName}.md`}
```
MikeLuDev, line 38: Not in scope for this PR, but should this url be hard coded? It seems that this limits this component to only `.md` files stored there.

If these files are part of the `members` source code, why not just pull them locally instead of remotely? This seems like an anti-pattern.

MikeLuDev (approved):
Really nice conversion to use hooks. 👍 

One question about a component, but does not affect this PR

src/shared-components/MarkdownFileRenderer.tsx
wdoug: We are getting them locally (see the function: `getMarkdownHTML`). This link just pulls up a github editor page for that file so it can streamline edits from folks that might not otherwise be comfortable editing text in the code.

